PR #98, opened 2024-07-16
Adds conventional otel resource attributes to the defaults

joe-elliott:
Adds OTEL resource attributes to the default list. Code comments indicate where each set of attribute names is sourced from.

This will improve the experience for those with conventional attribute names


ifrost:
Moving it back to To Do as we need to figure out how to display attributes with long names first https://github.com/grafana/explore-traces/issues/248

knylander-grafana:
This seems like a nice thing to mention in the docs as well.

joey-grafana:
@adrapereira you were working on a responsive group by right? If that's done then this would be a nice addition

PR #236, opened 2024-09-23
Consider all series when finding the max y axis

joe-elliott:
Fixes https://github.com/grafana/explore-traces/issues/213

Will take advice if there's a better way to do this, but this solves this issue:


joe-elliott:
So I could very much be wrong on what this change does, but to explain my reasoning.

image

It seems that as is the code only considers the values in `fields[1]` which is the error series. This should check `fields[1]` through `fields[N]`? and therefore check the values in the unset series and any others that may be present?

PR #242, opened 2024-09-25
Improve duration flow

adrapereira:
Don't mind the bug in the video where the Comparison tab is showing something else, that's been fixed.

https://github.com/user-attachments/assets/b2c9145f-5dac-4982-8aa3-bcf9434299bc

### DIFF
--- a/src/components/Explore/TracesByService/DurationComparisonControl.tsx
+++ b/src/components/Explore/TracesByService/DurationComparisonControl.tsx
@@ -36,7 +36,7 @@ export class DurationComparisonControl extends SceneObjectBase<ComparisonControl
 
     const isDisabled = selection?.type === 'auto';
     const tooltip = isDisabled
-      ? 'Slowest traces are already selected, navigate to the Comparison or Slow Traces tab for more details.'
+      ? 'Slowest traces are selected, navigate to the Comparison or Slow Traces tab for more details.'
       : undefined;
 
     return (


src/components/Explore/TracesByService/DurationComparisonControl.tsx
```diff
@@ -33,21 +36,27 @@ export class ComparisonControl extends SceneObjectBase<ComparisonControlState> {
...
 
-  public static Component = ({ model }: SceneComponentProps<ComparisonControl>) => {
-    const { buttonLabel } = model.useState();
+  public static Component = ({ model }: SceneComponentProps<DurationComparisonControl>) => {
     const { selection } = getTraceByServiceScene(model).useState();
     const styles = useStyles2(getStyles);
 
+    const isDisabled = selection?.type === 'auto';
+    const tooltip = isDisabled
+      ? 'Slowest traces are already selected, navigate to the Comparison or Slow Traces tab for more details.'
```
joey-grafana, line 45:
```suggestion
      ? 'Slowest traces are selected, navigate to the Comparison or Slow Traces tab for more details.'
```

src/components/Explore/TracesByService/REDPanel.tsx
```diff
@@ -193,6 +206,10 @@ export class REDPanel extends SceneObjectBase<RateMetricsPanelState> {
   }
 
   private buildSelectionAnnotation(state: TraceSceneState) {
+    if (!state.actionView || !['traceList', 'comparison'].includes(state.actionView)) {
```
joey-grafana, line 209: We should really start using consts here and in the other places this is used.

src/components/Explore/TracesByService/TracesByServiceScene.tsx
```diff
@@ -73,36 +73,42 @@ export class TracesByServiceScene extends SceneObjectBase<TraceSceneState> {
...
 
-    this.subscribeToState((newState, prevState) => {
-      const timeRange = sceneGraph.getTimeRange(this);
-      const selectionFrom = newState.selection?.timeRange?.from;
-      // clear selection if it's out of time range
-      if (selectionFrom && selectionFrom < timeRange.state.value.from.unix()) {
-        this.setState({ selection: undefined });
-      }
+    this._subs.add(
+      this.subscribeToState((newState, prevState) => {
```
joey-grafana, line 90: There may be an obvious answer here but why do we have to `this.subscribeToState` and add it to `this._subs.add`?

joey-grafana (changes requested):
Overall I like that you reintroduced the button as a way to clarify what the user can do or what the UI is already doing.

However, we're trying to select the slowest spans for our users but only apply this to the comparison/slow traces tab and this is creating some strange behaviour that the user will not understand.

When I select an area for comparison -> go to the breakdown tab -> click `Select slowest traces`, the two same queries for the breakdown tab will run again (of which one is red because a second slightly different query is run so soon after the first) and the view will refresh. Clicking that button seems to do nothing to the user until they go to the comparison/slow traces tab. Should we just take them there if they click the button?

https://github.com/user-attachments/assets/9ddd6e3b-49c8-4980-b0aa-4c37302dc990

The caveat is that the default tab is breakdown and the slowest traces are already selected. We have an option here to not select slowest traces by default, but only when comparison/slow tab is selected. There are side effects, but it simplifies the UX.

Screenshot 2024-09-26 at 11 27 10


@nadinevehling will have some good ideas here for sure.

joey-grafana:
> The two queries are unrelated to the slowest traces actually, they're related to the group by selector, it runs the first query using All and then a second using the value we automatically set to the variable (resource.service.name for example), we should still fix it though.

I do realise that's how it works now but it is unnecessarily refreshing the view which makes the user think something has changed but really it's only in the comparison/slow traces tab. On top of the extra queries like you mentioned. 

Let's either fix it here or open an issue so we don't forget please.

> It should automatically move you to the Comparison tab when you select an area on the histogram or click the "Select slowest traces" button.

If I press the "Select slowest traces" button on the breakdown tab it does not move me to the comparison tab.

> Let's try making the Comparison tab the default, it's what provides the most value to the user and helps reduce doubts around the slowest traces selection. We can roll back if we're not happy with the flow.

This would help fix the confusion above on initial load but with the slowest traces being selected. However, this tab takes a while longer to load than the breakdown -> are we sure users won't get frustrated by the longer wait?

joey-grafana:
> The button is disabled if the slowest traces are already selected (the default), are you clicking the disabled button?

In the last step, I can click the `Select slowest traces` button, but the tab stays on the breakdown, and still runs two more queries (of which one fails). The view is refreshing, but I'm staying on the same tab.

https://github.com/user-attachments/assets/e454061e-0c5e-45e1-b919-2876a624be31

joey-grafana (approved):
LGTM 🚀

PR #252, opened 2024-10-16
fix(Search): Make free text search case insensitive

grafakus:
### Description 

**Related issue** resolves https://github.com/grafana/explore-traces/issues/250

### Summary of the changes

In order to prevent some results to be absent from the free text search, this PR makes the search case-insensitive.

Idea for the future: adding a text case toggle, like we have in the Explore Logs app

image

### How to test

- The new unit tests should pass
- Manually, by reproducing the context described in the related issue

### DIFF
--- a/src/components/Explore/helpers/doesQueryMatchDataFrameLabels.ts
+++ b/src/components/Explore/helpers/doesQueryMatchDataFrameLabels.ts
@@ -1,0 +1,12 @@
+import { DataFrame } from '@grafana/data';
+
+export const doesQueryMatchDataFrameLabels = (searchQuery?: string) => (dataFrame: DataFrame) => {
+  const pattern = searchQuery?.trim();
+  if (!pattern) {
+    return true;
+  }
+
+  const regex = new RegExp(pattern, 'i');
+
+  return dataFrame.fields.some((f) => (!f.labels ? false : Object.values(f.labels).find((label) => regex.test(label))));
+};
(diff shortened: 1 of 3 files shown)


src/components/Explore/helpers/doesQueryMatchDataFrameLabels.ts
```diff
@@ -0,0 +1,12 @@
+import { DataFrame } from '@grafana/data';
+
+export const doesQueryMatchDataFrameLabels = (searchQuery?: string) => (dataFrame: DataFrame) => {
+  const pattern = searchQuery?.trim();
+  if (!pattern) {
+    return true;
+  }
+
+  const regex = new RegExp(pattern, 'i');
```
ifrost, line 9: That would allow to search using regular expression which is kind of cool, but I was wonder if:
- should we be more explicit about it (e.g. in the prompt) + add tests coverage for it?, or
- or lowercase label+input and do simple case-insensitive search?

ifrost (commented):
LGTM 👍  just a question about regex vs case-insensitive search

src/components/Explore/helpers/doesQueryMatchDataFrameLabels.ts
grafakus: Good question, I believe it all depends how we want this feature to evolve. IMHO, I'd love to see all apps going towards what (e.g.) VSCode provides:

image

I guess for now, the PR is good enough and we can discuss about it in the future.

PR #260, opened 2024-10-30
Add a bunch of new unit tests

joey-grafana:


src/components/Explore/actions/AddToFiltersAction.test.tsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react'; 
+import { AddToFiltersAction, addToFilters } from './AddToFiltersAction';
+import { DataFrame } from '@grafana/data';
+import { AdHocFiltersVariable } from '@grafana/scenes';
+
+jest.mock('../../../utils/utils', () => ({
```
grafakus, line 7:
```suggestion
import { getFiltersVariable, getLabelValue } from '../../../utils/utils';

jest.mock('../../../utils/utils');
```

src/components/Explore/actions/AddToFiltersAction.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+  getFiltersVariable: jest.fn(),
+  getLabelValue: jest.fn(),
+}));
+
+const mockGetFiltersVariable = require('../../../utils/utils').getFiltersVariable;
+const mockGetLabelValue = require('../../../utils/utils').getLabelValue;
+
+describe('AddToFiltersAction', () => {
+  let variable: AdHocFiltersVariable;
```
grafakus, line 16: **[non-blocking]**: I personally find it more difficult to maintain tests in the long term when they use closures instead of factory functions.

For instance, something similar to:

```js
function buildAction(labels: Record<string, string> = { label1: 'value1', label2: 'value2' }) {
  const frame = {
    fields: [{ labels }],
  } as unknown as DataFrame;

  const setVariableState = jest.fn();
  const onClick = jest.fn();

  (getLabelValue as jest.Mock).mockReturnValue('value1');

  (getFiltersVariable as jest.Mock).mockReturnValue({
    state: { filters: [] },
    setState: setVariableState,
  } as unknown as AdHocFiltersVariable);

  return {
    action: new AddToFiltersAction({ frame, onClick }),
    onClick,
    setVariableState,
  };
}
```

would make the tests fairly easy to understand at first glance:

```js
  it('should add filter when no labelKey and exactly one label exists', () => {
    /// arrange
    const { action, onClick, setVariableState } = buildAction({ label1: 'value1' });

    // act
    action.onClick();

    // assert
    expect(setVariableState).toHaveBeenCalledWith({
      filters: [{ key: 'label1', operator: '=', value: 'value1' }],
    });
    expect(onClick).toHaveBeenCalledWith({ labelName: 'label1' });
  });
```

What do you think?

jest.config.js
```diff
@@ -7,6 +7,6 @@ module.exports = {
   ...require('./.config/jest.config'),
   moduleNameMapper: {
```
grafakus, line 8: Maybe we could aslo add:

```js
  resetMocks: true,
  clearMocks: true,
  resetModules: true,
```

Is there a situation where we don't want these options to be true?

src/components/Explore/filters/AddFilter.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    expect(screen.getByText('Filter by attribute...')).toBeInTheDocument();
+  });
+
+  it('renders add filter bar without text when filters are present', () => {
+    render(<AddFilter model={mockModel as unknown as FilterByVariable} otherFiltersLength={1} />);
+    expect(screen.queryByText('Filter by attribute...')).not.toBeInTheDocument();
+  });
+
+  it('renders FilterRenderer when _wip is defined', () => {
+    const mockModel = {
```
grafakus, line 38: Would it be possible to pass a real instance as the model? Something similar to:

```js
const model = new FilterByVariable({ ... })

jest.spyOn(model, 'useState').mockReturnValue({ _wip: { key: 'testKey' } });
```

src/components/Explore/filters/FilterRenderer.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+    await act(async () => {
+      render(<FilterRenderer filter={mockFilter} model={mockModel as unknown as FilterByVariable} />);
+    });
+    
+    await user.click(screen.getByText('testKey'));
+    expect(screen.getByText('kind')).toBeInTheDocument();
+    expect(screen.getByText('resource.tag1')).toBeInTheDocument();
+    await user.click(screen.getByText('kind'));
+    
+    expect(mockModel._updateFilter).toHaveBeenCalledWith(mockFilter, { key: 'kind' });
```
grafakus, line 73: Is it the only expectation we can have? As `model._updateFilter` is a "private" method from a 3rd-party library, could we assert on something that is in our control? Something we can observe in the document for instance?
(I'm not sure because I'm not super familiar with the code)

My concern is that if Scenes `AdHocFiltersVariable` is modified, this test will fail, and it shouldn't.

src/utils/dates.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+describe('formatDuration', () => {
+  it('formats durations less than a second correctly', () => {
+    expect(formatDuration(500)).toBe('500μs');    // 500 microseconds
+    expect(formatDuration(5000)).toBe('5ms');     // 5000 microseconds
+    expect(formatDuration(1000000)).toBe('1s');   // 1,000,000 microseconds
+  });
+
+  it('formats durations in seconds and minutes', () => {
+    expect(formatDuration(1500000)).toBe('1.5s');    // 1,500,000 microseconds
```
grafakus, line 18: **[nit]** Could we use numeric separators? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators

src/utils/dates.test.ts
```diff
@@ -0,0 +1,78 @@
+import { sceneGraph, SceneObject } from '@grafana/scenes';
+import { formatDuration, getStepForTimeRange, ONE_DAY, ONE_HOUR, ONE_MILLISECOND, ONE_MINUTE, ONE_SECOND } from './dates';
+
+jest.mock('@grafana/scenes', () => ({
+  sceneGraph: {
+    getTimeRange: jest.fn(),
+  },
+}));
```
grafakus, line 8:
```suggestion
jest.mock('@grafana/scenes');
```

src/components/Explore/actions/AddToFiltersAction.test.tsx
joey-grafana: Nice spot, they're actually not needed in the mock.

src/components/Explore/actions/AddToFiltersAction.test.tsx
joey-grafana: Honestly I don't have a preference. I like the organisation of a factory function however I also like the clarity of the various bits laid out in the `beforeEach`.

jest.config.js
joey-grafana: I think this promotes a cleaner approach to the tests. Will add it!

src/utils/dates.test.ts
joey-grafana: Absolutely

src/components/Explore/filters/AddFilter.test.tsx
joey-grafana: Sure, can do!

src/components/Explore/filters/FilterRenderer.test.tsx
joey-grafana: The problem here is that I need to mock parts of the model, because creating the variable and letting the code flow as it normally would, will create issues as it flows through scenes and expects various parts of the model to be defined (and which will not update the UI correctly unless it is defined correctly), which it is not as this does not reflect the purpose of the test and requires extra setup to work correctly.

IMO, it's better to mock high level functions, where you have more control over the flow and what's returned. Then you can check that the code flows in an expected way and test accordingly.

I agree that we could test something in the document instead of testing a method outside of the app, but that actually requires not mocking the `_updateFilter` method for example, which is an issue for the reason mentioned above. I have however update this to use `jest.spyOn` for the methods I want to watch rather than mock the model itself.

In any case, I don't see a huge amount of value in not testing this method from scenes is called with the correct data as it's what we call directly in our `onChange`. Now I could add a util function that wraps `_updateFilter` in our `onChange` and test that is called with the correct input, but that's really just adding a layer of misdirection for the sake of the tests.

grafakus (approved):

PR #262, opened 2024-11-05
Add e2e tests and add to CI

joey-grafana:
Adds several e2e tests to get us started. Also adds them to the CI for every PR.


.github/actions/build/action.yml
```diff
@@ -26,7 +26,7 @@ runs:
         cache: 'npm'
 
     - name: Install dependencies
-      run: npm ci
```
grafakus, line 29: Why did you change it?

.github/actions/build/action.yml
```diff
@@ -48,6 +48,30 @@ runs:
...
       shell: bash
+    
+    - name: Start Grafana Docker container
+      run: npm run server:ci
+      shell: bash
+    - name: Install Playwright Browsers
+      run: npx playwright install chromium --with-deps
+      shell: bash
+    - name: Wait for Docker container
+      run: sleep 60
```
grafakus, line 59: Why not just probing that the app is ready? Using https://github.com/marketplace/actions/wait-for-http-responses or a similar approach?

e2e/components.spec.ts
```diff
@@ -0,0 +1,61 @@
+import { expect, test } from '@grafana/plugin-e2e';
+import { ExplorePage } from './fixtures/explore';
+
+test.describe('components', () => {
+  let explorePage: ExplorePage;
+  
+  test.beforeEach(async ({ page }) => {
+    explorePage = new ExplorePage(page);
```
grafakus, line 8: Why not using [Playwright fixtures](https://playwright.dev/docs/test-fixtures)?

grafakus (commented):
Nice start! To prevent flakiness, my recommendations would be to:
- Set a fixed version for Grafana, Tempo and any other service that the app depends upon, always using the `latest` version can make a new execution behave differently 
- Execute the E2E tests inside a Docker container, Playwright provides [an official image](https://playwright.dev/docs/docker), it will also help when using screenshots in the tests (the OS/browser used to generate the screenshots will always be the same)
- Keep in mind the philosophy of E2E testing: testing _from the point of view of the user_. In this sense, it's better to assert visual elements (what the user notice and see happening in the UI) than (e.g.) what the URL looks like (in `navigation.spec.ts`) 

I also found valuable to separate the execution of the E2E tests in a separate GtHub action job. The checks on the PR are clearer and it can also pave the way for parallelizing the execution of (e.g.) unit tests vs E2E tests.

package.json
```diff
@@ -5,14 +5,18 @@
   "scripts": {
     "build": "webpack -c ./webpack.config.ts --env production",
     "dev": "webpack -w -c ./webpack.config.ts --env development",
+    "e2e": "playwright test -x --trace on",
```
grafakus, line 8: To improve the DevX, we should add the [--ui flag](https://playwright.dev/docs/test-ui-mode)

playwright.config.ts
```diff
@@ -0,0 +1,52 @@
...
+  expect: { timeout: 15000 },
+
+  /* Configure projects for major browsers */
+  projects: [
+    // 1. Login to Grafana and store the cookie on disk for use in other tests.
+    {
+      name: 'auth',
+      testDir: pluginE2eAuth,
+      testMatch: [/.*\.js/],
+    },
```
grafakus, line 44:
```suggestion
```

playwright.config.ts
```diff
@@ -0,0 +1,52 @@
...
+    // 2. Run tests in Google Chrome. Every test will start authenticated as admin user.
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'], storageState: 'playwright/.auth/admin.json' },
+      dependencies: ['auth'],
```
grafakus, line 49:
```suggestion
```

Auth is not necessary

playwright.config.ts
```diff
@@ -0,0 +1,52 @@
...
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: 'html',
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: 'http://localhost:3001',
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
```
grafakus, line 28: If not configured by default, it's worth saving a video on first retry and taking screenshot on failure.

e2e/components.spec.ts
joey-grafana: Feels cleaner to define the url once in the ExplorePage class and call the method. Is there another benefit to using Playwright fixtures here?

package.json
joey-grafana: Nice 👍

joey-grafana:
> Set a fixed version for Grafana, Tempo and any other service that the app depends upon, always using the latest version can make a new execution behave differently

Update
We discussed this in standup and will be going with a fixed version for now!

Before standup
It was deliberate setting this to latest because I'd rather a failing test telling us something is off with the API rather than it failing silently. Having said that, we should probably catch issues with the image in development (as long as we regularly update our images). I'll change it to a particular version for now until at least we're sure that the e2e tests are stable.

> Execute the E2E tests inside a Docker container, Playwright provides [an official image](https://playwright.dev/docs/docker), it will also help when using screenshots in the tests (the OS/browser used to generate the screenshots will always be the same)

Will look into this, either this PR or perhaps in a follow up PR as an improvement.

> Keep in mind the philosophy of E2E testing: testing from the point of view of the user. In this sense, it's better to assert visual elements (what the user notice and see happening in the UI) than (e.g.) what the URL looks like (in navigation.spec.ts)

Yes, good point. This is the general purpose of these tests in the PR, however our state is essentially stored in the URL so there is a lot of value (covers a lot of UI decisions) to add in a small test to make sure the default URL is correct which can find a lot of issues before they occur in the UI. This may be the exception that proves the rule for Explore Traces!

> I also found valuable to launch the execution of the E2E tests in a separate GitHub action job. The checks on the PR are clearer and it can also pave the way for parallelizing the execution of (e.g.) unit tests vs E2E tests.

Nice, this feels cleaner! The only issue here is that if they're running separately does that mean that even if the `end-to-end` job fails, the build artefact will still be uploaded?

e2e/components.spec.ts
grafakus: Yes, by extending the [base test](https://playwright.dev/docs/test-fixtures#with-fixtures) (sorry I didn't link to the paragraph I had in mind), each test can received the `explorePage` as a param of the callback. For instance, like in [this suite of tests](https://github.com/grafana/explore-profiles/blob/main/e2e/tests/ad-hoc-view/ad-hoc-view.spec.ts).

You can pretty much [pass anything](https://github.com/grafana/explore-profiles/blob/main/e2e/fixtures/index.ts) that can help writing tests in a simpler way.

playwright.config.ts
grafakus: I meant adding [this config](https://github.com/grafana/explore-profiles/blob/main/e2e/config/playwright.config.common.ts#L37-L41). But I'm unsure if, by default, it's already configured.

grafakus:
> Yes, good point. This is the general purpose of these tests in the PR, however our state is essentially stored in the URL so there is a lot of value (covers a lot of UI decisions) to add in a small test to make sure the default URL is correct which can find a lot of issues before they occur in the UI. 

Then... wouldn't it be less costly/faster to write a unit test for it?

grafakus:
> Nice, this feels cleaner! The only issue here is that if they're running separately does that mean that even if the end-to-end job fails, the build artefact will still be uploaded?

Yes, they would be uploaded to GitHub, for a few days (5? in the case of Explore Traces).

We can postpone the parallelization, but at least if we have separate jobs, the build checks would be easier to understand.

.github/actions/build/action.yml
```diff
@@ -48,6 +48,34 @@ runs:
...
+      run: npm run server:ci
+      shell: bash
+    - name: Install Playwright Browsers
+      run: npx playwright install chromium --with-deps
+      shell: bash
+    - name: Wait for Docker container
+      uses: cygnetdigital/wait_for_response@v2.0.0
+      with:
+        url: 'http://localhost:3001/'
+        responseCode: '200,500'
```
grafakus, line 62: Why HTTP 500?

grafakus (approved):
Feel free to address more suggestions. It's good base and we can keep on improving as time passes!

PR #264, opened 2024-11-11
Improve functionality when opened inside sidecar

aocenas:
Main things being changed here:
1. Opening a single trace correctly references context locationService instead of the global singleton which prevented the navigation inside sidecar
2. The detail view (like the trace view) is opened as main view, not the split view when inside the sidecar. This is to prevent triple split view in that case.
3. Small responsive update to the main controls so they wrap around when there isn't enough space


src/pages/Explore/TraceExploration.tsx
```diff
@@ -196,11 +204,30 @@ export class TraceExploration extends SceneObjectBase<TraceExplorationState> {
...
 
   static Component = ({ model }: SceneComponentProps<TraceExploration>) => {
-    const { body } = model.useState();
+    const { singleBody, splitBody } = model.useState();
     const styles = useStyles2(getStyles);
 
+    // The API looks a bit weird because duh we are opened if we are here, but this specifically means we are in a
+    // sidecar with some other app. In that case we don't want to show additional split layout as there is not much
+    // space and 3 splits is a bit too much.
+    const body: SceneObject = sidecarServiceSingleton_EXPERIMENTAL?.isAppOpened(pluginJson.id) ? singleBody : splitBody;
```
joey-grafana, line 213: Can we also check the feature toggle is enabled?

aocenas:
@joey-grafana 
> View linked span on the service structure tab does not work in sidecar.

Good catch, fixed. 

> Opening up Explore Traces in sidecar still creates an invalid query.

So not really going to fix it here. Like the discussion about how to normalize labels between apps/backend is bit complicated. Metrics for example have a whole project about how to mat their label to logs so this is a bit out of my domain knowledge right now and will have to be figured out later on.

> When sidecar is opened, the Related traces button is missing. If the user wants to update the context in sidecar because they have selected a new label in Loki they cannot.

So at first I was thinking about this as just "Open" button so it seems reasonable to not show it if Traces are already opened but as you suggest this actually works for sort of state update. So will leave it visible for now, in the future though I assume the button will be in some submenu (cause there would related metrics and profiles) so at that point we will have to figure out some better way to sync labels after changes.

> The top padding on the need help and time range buttons are reduced when in sidecar

I actually don't see it locally nor in ops so not sure what is going on. Don't think the sidecar itself should change any styles of the app inside it. Do you see it in ops?

> Can we add a tooltip to the x in the sidecar view that says something like Close sidecar. Multiple close buttons which are near each other are a bit confusing from a UX perspective.

Yes but that is part of Grafana code so will have to do it there later.

> Copying the URL via the share button opens the same view twice, not sure what the best approach here is but this can be improved. Sidecars context is not in the URL. ...

So yeah this is something I am working on and figuring out exact rules. In the long run I think there would be 2 kinds of apps one that will have state in the URL like this but also ones that don't like the Explorations app which does not make sense to be shared in the URL. Having state of sidecar in the URL just isn't implemented yet so this uses the non URL mode but hope to change that soon. But it's also something to be changed in the Grafana code.

joey-grafana:
>I actually don't see it locally nor in ops so not sure what is going on. Don't think the sidecar itself should change any styles of the app inside it. Do you see it in ops?

Yes, also in ops.

Screenshot 2024-11-21 at 09 24 30


> So not really going to fix it here. Like the discussion about how to normalize labels between apps/backend is bit complicated. Metrics for example have a whole project about how to mat their label to logs so this is a bit out of my domain knowledge right now and will have to be figured out later on.

Tbh this feels like a bad experience to open sidecar with a broken query. Are we including the query to show that context can be passed between apps (related traces :D)? Can you just open sidecar without the query until we figure out how to translate the labels, as in what is the benefit of sending the query right now? In any case, when you do get around to normalizing the labels, [importFromAbstractQueries](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/datasource.ts#L507) should offer insight.

Btw, the build is failing please take a look.

aocenas:
> Tbh this feels like a bad experience to open sidecar with a broken query.

As discussed on the call, this is more of a showcase of the capabilities than a finalized product so yeah that part is naive and probably does not work many times (only if the labeling is the same). How the normalization of labels across signals should work so that we can correlate data between is something we will have to figure out if we go this way (well even without a sidecar it is something to figure out for any other kind of correlation). Also it's more of an app domain thing and we will be talking about it next week how to approach it.

joey-grafana (approved):

PR #278, opened 2024-12-17
Exemplars support and misc fixes

### LINKED ISSUE #73: Drawing a vertical box on compare gives an error
If you draw a vertical box then the app will send the same start/end time. Perhaps it should just increment end by 1 if they are exactly equal?

image

### LINKED ISSUE #95: App repeatedly runs bad queries
When adding filters, clearing investigations or making other updates on the "second page" of the app it repeatedly executes pairs of queriers. One of those queries always fails with a parse error.

image

The first/failing query always looks something like:

```
{nestedSetParent<0&&resource.service.name="CockroachDB" && All != nil} | rate() by(All, status)
```

The "All" keyword is not valid Traceql. The app otherwise seems to be working fine so I believe we can just remove this unnecessary/incorrect call.

adrapereira:
Fixes #95 
Fixes #73

### DIFF
--- a/src/utils/exemplars.ts
+++ b/src/utils/exemplars.ts
@@ -1,13 +1,12 @@
 import { map, Observable } from 'rxjs';
 import { DataFrame, DataTopic, Field } from '@grafana/data';
-import { CustomTransformerDefinition, SceneObject } from '@grafana/scenes';
-import { getTraceExplorationScene } from './utils';
+import { CustomTransformerDefinition } from '@grafana/scenes';
+import { LocationService } from '@grafana/runtime';
 
-export const exemplarsTransformations = (scene: SceneObject): CustomTransformerDefinition[] => [
+export const exemplarsTransformations = (locationService: LocationService): CustomTransformerDefinition[] => [
   {
     topic: DataTopic.Annotations,
     operator: () => (source: Observable<DataFrame[]>) => {
-      const traceExplorationScene = getTraceExplorationScene(scene);
       return source.pipe(
         map((data: DataFrame[]) => {
           return data.map((frame) => {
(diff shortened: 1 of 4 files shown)


grafakus (approved):
👍🏾👍🏾 LGTM

Just a small suggestion (or question), what do you think about:

```suggestion
export const exemplarsTransformations = (locationService: LocationService): CustomTransformerDefinition[] => [
```
?

By limiting what we pass to what's used by the function:
- we make the dependency of `exemplarsTransformations` to the location service explicit (easier to reason about)
- we enable flexibility, if we want to move the location service out of `TraceExploration`
- so, we also simplify testing

No changes required, the code looks good in general!

src/utils/exemplars.ts
```diff
@@ -0,0 +1,42 @@
...
+                  {
+                    title: 'View trace',
+                    url: '#${__value.raw}',
+                    onClick: (event) => {
+                      event.e.stopPropagation(); // Prevent the click event from propagating to the parent anchor
+                      const parentAnchorHref = event.e.target.parentElement.parentElement.href;
+                      const traceId = parentAnchorHref.split('#')[1];
+                      traceExplorationScene.state.locationService.partial({
+                        traceId,
+                      });
```
joey-grafana, line 29: Can we be a bit more careful here please. Events/parentElement/targets are notorious for being undefined or unexpected. A few optional chaining ?'s would be safer so we don't get any potential runtime errors, especially if changes to the DOM/parent structure occur. Same with the `split('#')[1]` in case there are not even two parts to access the second.

We're also assuming `parentAnchorHref.split('#')[1]` will always be the traceId, but we don't check the formatting to be sure it's a traceId. We don't want to update the state with an invalid traceid!

Adding a fallback so that we link to Classic Explore if any of the above is a problem would be ideal but not required.

joey-grafana (changes requested):
Thanks for the fixes 👍 

There are still some cancelled query errors when duration is selected but this is an improvement overall.

joey-grafana (approved):
Nice work! Just reviewing the other PR now 💯

PR #279, opened 2024-12-19
Explore Traces Homepage

joey-grafana:
In this PR we have added a home page for Explore Traces! 🥳 🚀 

The homepage allows you to get a quick overview of your services without the need to add filters or select your desired RED metric.

Through this view you can easily see your services with errors and also your slowest services: two key areas that we constantly strive to monitor.

Keep in mind this is only a v1 for the homepage, there is much more planned! Thanks @nadinevehling for all your work!

Screenshot 2024-12-19 at 09 26 16


alexbikfalvi:
Great work, leaving here my comments that we discussed.

- For **Errored services**, the the list of traces  is not always the same when refreshing the page. Is it possible to be consistent and show always the last 10?
- **Trace name** might confuse users on what is being shown (service name and resource name), as users have mentioned that sometimes they struggle to know what's being displayed. Would it make sense to be more specific (say "Service name" and "??? name")?
- When there is no Tempo data source we show the following ("Errored services" has a "Data source error" and "Slow services" is missing, which looks like a bug). As a low effort fix, we could customize the error message (e.g. "There are no Tempo data sources" instead of "Data source error").
Screenshot 2025-01-02 at 3 38 33 PM

joey-grafana:
> For Errored services, the the list of traces is not always the same when refreshing the page. Is it possible to be consistent and show always the last 10?

I think we will need the topk API for getting the top erroring services @joe-elliott @mdisibio (something that we add frontend support for in the next iteration).

> Trace name might confuse users on what is being shown (service name and resource name), as users have mentioned that sometimes they struggle to know what's being displayed. Would it make sense to be more specific (say "Service name" and "??? name")?

Service name probably makes more sense here.

> When there is no Tempo data source we show the following ("Errored services" has a "Data source error" and "Slow services" is missing, which looks like a bug). As a low effort fix, we could customize the error message (e.g. "There are no Tempo data sources" instead of "Data source error").

Yes I can update this.

joey-grafana:
For a query like `{nestedSetParent < 0 && status = error} | count_over_time() by (resource.service.name, name, trace:id)` there are thousands of frame results for instant metrics (what's the best way to limit this?). We could count the errors, but they're often 0 (as root span apparently not counted, only child spans of root), and then sort by errors but each trace always has 0 or 1 errors (with local & dev - not sure how often in prod there are more than one error). 

I wonder would the best move be to group by `span:id` as well and then combine all errors from each trace.

src/components/Home/DurationAttributePanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { yBucketToDuration } from 'components/Explore/panels/histogram';
+import { AttributePanel, SkeletonComponent } from './AttributePanel';
+import { getErrorMessage, getNoDataMessage } from 'utils/utils';
+import { AttributePanelScene } from './AttributePanelScene';
+
+export interface DurationAttributePanelState extends SceneObjectState {
+  panel?: SceneFlexLayout;
+}
+
+export class DurationAttributePanel extends SceneObjectBase<DurationAttributePanelState> {
```
ifrost, line 24: Looks very similar to `AttributePanel` - would it make sense to extend the class to reuse some code? 🤔

src/utils/utils.ts
```diff
@@ -43,11 +48,26 @@ export function newTracesExploration(
...
+    initialDS,
+    $timeRange: new SceneTimeRange({ from: 'now-30m', to: 'now' }),
+  });
+}
+
+export function getErrorMessage(data: SceneDataState) {
+  return data?.data?.error?.message ?? 'There are no Tempo data sources';
+}
+
+export function getNoDataMessage(context: string) {
```
ifrost, line 67: When hovering over the icon it shows a pointer cursors but it's not interactive

copy

joey-grafana:
Thanks for all the suggestions (which should be covered in the last commits). This is what the UI looks like.

Screenshot 2025-01-08 at 14 23 40

src/components/Home/AttributePanel.tsx
```diff
@@ -55,17 +60,45 @@ export class AttributePanel extends SceneObjectBase<AttributePanelState> {
...
+                  }
+
+                  const minDuration = yBucketToDuration(minBucket - 1, yBuckets);
+                
+                  this.setState({
+                    panel: new SceneFlexLayout({
+                      children: [
+                        new AttributePanel({ 
+                          query: {
+                            query: `{nestedSetParent<0 && kind=server && duration > ${minDuration}} | by (resource.service.name)`,
```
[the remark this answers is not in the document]
joey-grafana: Updated the query 👍 

> Also, this will suffer of the same issues as the errors table had, the results won't be consistent across refreshes since it's a search query.

Technically, yes. But I've just tested it with dev and after 10 refreshes it was always the same. Our query is something like `{nestedSetParent<0 && kind=server && duration > 9s}` depending what's in the slowest percentile and the results are also then ordered by slowest - there's not that many results. Also we have already discussed this, that it's valuable to link directly to the slow trace from this panel. We can adjust in the future if it becomes something the user really wants.

src/components/Home/AttributePanelRows.tsx
```diff
@@ -0,0 +1,171 @@
...
+          'var-metric': type,
+        }
+        const url = urlUtil.renderUrl(EXPLORATIONS_ROUTE, params);
+        return `${url}&var-filters=nestedSetParent|<|0`;
+      }
+
+      const getTotalErrs = (df: DataFrame) => {
+        const valuesField = df.fields.find((f) => f.name !== 'time');
+        return valuesField?.values?.reduce((x, acc) => {
+          if (typeof x === 'number' && !isNaN(x)) {
```
alexbikfalvi, line 54: (as discussed in Slack): Should we check here also that `acc` is a number? If `x` is a number but `acc` is NaN, the result of `x + acc` is NaN.

joey-grafana: I've set acc to 0 by default, so that ensure it's a number and we do not have to check if acc is a number as we already check if x is a number and only add it to x if it is.

PR #291, opened 2025-01-14
Hide span details in structure tab

joey-grafana:
Fixes https://github.com/grafana/explore-traces/issues/253

### DIFF
--- a/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
+++ b/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
@@ -393,6 +393,21 @@ const getStyles = (theme: GrafanaTheme2) => {
       'div[data-testid="span-detail-component"] > :nth-child(4) > :nth-child(1)': {
         display: 'none',
       },
+
+      // Hide span details row
+      '.span-detail-row': {
+        display: 'none',
+      },
+
+      // Remove cursor pointer as soan details is hidden
+      'div[data-testid="TimelineRowCell"]': {
+        'button[role="switch"]': {
+          cursor: 'text',
+        }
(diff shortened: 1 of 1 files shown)


src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
```diff
@@ -393,6 +393,21 @@ const getStyles = (theme: GrafanaTheme2) => {
       'div[data-testid="span-detail-component"] > :nth-child(4) > :nth-child(1)': {
         display: 'none',
       },
+
+      // Hide span details row
+      '.span-detail-row': {
```
ifrost, line 398: Where's that class being used? 🤔  Couldn't find reference to it anywhere.

Hiding with CSS may break if class names change in the TraceView so if we want to go this route I think we should have a test that would ensure elements are really hidden. Also, wouldn't it be nicer if we added a new property to TraceView to control it?

joey-grafana: I've added it to the trace view in the [Grafana repo](https://github.com/grafana/grafana/pull/98946) PR.

I've also had a look at creating a test for this e.g. render the Traces panel and make sure that some of the child components of the span detail are not visible but it's kind of tricky to mock / provide dummy data for the Structure Tab :( do you have any suggestions?

Adding a new property is a nice touch and was almost going to do that but thought it's a bit overkill for this. We already hide the minimap, header, span and resource accordions via styling in Explore Traces and wanted to hide span details in the same way rather than introduce a new prop that will probably only be used by the app and may confuse users.

ifrost (approved):

PR #294, opened 2025-01-17
Add back and forward support for app actions

joey-grafana:
Fixes https://github.com/grafana/explore-traces/issues/34

### DIFF
--- a/src/components/Explore/panels/histogram.ts
+++ b/src/components/Explore/panels/histogram.ts
@@ -2,6 +2,7 @@ import { getTraceByServiceScene, shouldShowSelection } from '../../../utils/util
 import { ComparisonSelection } from '../../../utils/shared';
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from '../../../utils/analytics';
 import { PanelBuilders, SceneFlexItem, SceneFlexLayout, SceneObject } from '@grafana/scenes';
+import { locationService } from '@grafana/runtime';
 
 export function getHistogramVizPanel(scene: SceneObject, yBuckets: number[]) {
   const parent = getTraceByServiceScene(scene);
@@ -22,6 +23,7 @@ export function getHistogramVizPanel(scene: SceneObject, yBuckets: number[]) {
         const rawSelection = args[0];
         // @ts-ignore
         const newSelection: ComparisonSelection = { type: 'manual', raw: rawSelection };
+        locationService.partial({ 'selection': newSelection });
 
         newSelection.timeRange = {
           from: Math.round((rawSelection.x?.from || 0) / 1000),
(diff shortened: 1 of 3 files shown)


joey-grafana:
> Is manipulating the URL params manually the only option? It feels unintuitive and unsafe to me. I would guess that we can update the variables state or the scene state and let scenes automatically sync the state with the URL, propagating the changes as needed.

I thought it was already syncing the state with the url? I'm looking to add a historyStep and would love to have the url state handle that but I'm getting a bit confused with the differing scenes versions and the appropriate approach to accomplish this. I've seen an example in the [docs](https://grafana.com/developers/scenes/url-sync) of using the url-sync but also seen a [comment](https://github.com/grafana/explore-profiles/pull/128#discussion_r1728883187) that mentions using `locationService.partial` so not certain right now which approach is best.

src/components/Explore/panels/histogram.ts
```diff
@@ -37,7 +37,9 @@ export function getHistogramVizPanel(scene: SceneObject, yBuckets: number[]) {
         const yTo = yBucketToDuration(args[0].y?.to || 0, yBuckets);
         newSelection.duration = { from: yFrom, to: yTo };
 
-        parent.setState({ selection: newSelection });
+        parent._urlSync.performBrowserHistoryAction(() => {
```
ifrost, line 40: I think this should be rather encapsulated in a separate function in the parent like [an example in docs](https://github.com/grafana/scenes/blob/main/docusaurus/docs/url-sync.md#sceneobjecturlsyncconfig). This way we will not need expose _urlSync of the parent which should remain private.

cspell.config.json
```diff
@@ -48,6 +48,7 @@
     "gdev",
     "lokiexplore",
     "viewports",
+    "Descendents",
```
ifrost, line 51: Shouldn't this be **Descendants**, not [Descendents](https://en.wikipedia.org/wiki/Descendents)?

joey-grafana: Yes, but it's a common misspelling and I wanted to be [consistent](https://github.com/grafana/scenes/blob/33d096e023645c65fad2aa8ad2ba63c5968a5207/packages/scenes/src/core/sceneGraph/sceneGraph.ts#L257) :)

src/pages/Explore/TraceExploration.tsx
```diff
@@ -95,6 +96,15 @@ export class TraceExploration extends SceneObjectBase<TraceExplorationState> {
...
     });
+
+    const metricVariable = this.getMetricVariable();
+    this._subs.add(
+      metricVariable.subscribeToState((newState, prevState) => {
+        if (newState.value !== prevState.value) {
+          this.setState({ metric: newState.value as MetricFunction });
+        }
+      })
+    );
```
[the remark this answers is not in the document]
joey-grafana: I'm not sure it does tbh. Can you check out the branch and check if you remove metric from the scenes state it looks like the url sync does not work correctly?

PR #326, opened 2025-02-06
Add filter to homepage

joey-grafana:
- Adds a filter to the homepage which can update the query.
- Set to `resource.service.name` by default.
- If only attribute is set, it will be used in the `by` section of the query (for errors).
- If value is also set, `attribute=value` will be used in the `{ .. }` section of the query (for errors and duration).

Internal user interviews showed us that users would like the ability to select which service/cluster/etc that are most interested in or simply to see errors/slow traces for an attribute other than `resource.service.name`. Having spoken with @nadinevehling & @alexbikfalvi, we decided to allow users to add an optional filter for this iteration and based on feedback we will decide if we want to allow them to add more than one filter/create views (something like what Explore Logs does), but for this iteration it was requested to walk before we run :D 

Next enhancements;
- Remember the users last filter,
- Limit values in the dropdown

Screenshot 2025-02-06 at 15 36 55

### DIFF
--- a/src/components/Explore/filters/FilterByVariable.tsx
+++ b/src/components/Explore/filters/FilterByVariable.tsx
@@ -6,6 +6,7 @@ import { VariableHide } from '@grafana/schema';
 
 import { FilterSetRenderer } from './FilterSetRenderer';
 import { VAR_FILTERS, explorationDS } from 'utils/shared';
+import { isNumber } from 'utils/utils';
 
 export type FilterByVariableState = ConstructorParameters<typeof AdHocFiltersVariable>[0] & {
   initialFilters?: AdHocVariableFilter[];
@@ -35,7 +36,6 @@ export function renderTraceQLLabelFilters(filters: AdHocVariableFilter[]) {
   // and avoid invalid queries like '{ && key=value }'
   return expr.length ? expr : 'true';
 }
-const isNumber = /^-?\d+\.?\d*$/;
 
 function renderFilter(filter: AdHocVariableFilter) {
   let val = filter.value;
--- a/src/components/Home/AttributePanel.tsx
+++ b/src/components/Home/AttributePanel.tsx
@@ -28,6 +28,7 @@ export interface AttributePanelState extends SceneObjectState {
   title: string;
   type: MetricFunction;
   renderDurationPanel?: boolean;
+  filter?: string;
 }
 
 export class AttributePanel extends SceneObjectBase<AttributePanelState> {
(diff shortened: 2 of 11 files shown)


nadinevehling:
@joey-grafana Small feedback, I'd enlarge the spacing between the data source input and the text "filter" sightly to distinguish both filters. Right now, it's hard to tell which belongs together.

alexbikfalvi:
Fantastic work @joey-grafana - leaving a few individual comments below.

### Error for certain filters

There is an error for certain filters, such as `span:name = add-checkout` for instance.
Screenshot 2025-02-18 at 11 34 45 AM

The query is the following:
```
http://localhost:3000/api/datasources/proxy/uid/tempo/api/metrics/query_range?query=%7BnestedSetParent%20%3C%200%20%26%26%20status%20%3D%20error%20%26%26%20span%3Aname%3Dadd-checkout%7D%20%7C%20count_over_time()%20by%20(resource.service.name)&start=1739873095&end=1739874895&step=1800s
```
answered with 400 Bad Request - compiling query: parse error at line 1, col 53: syntax error: unexpected IDENTIFIER

This could be because of missing quotes around `add-checkout`, the following query works:
```
http://localhost:3000/api/datasources/proxy/uid/tempo/api/metrics/query_range?query=%7BnestedSetParent%20%3C%200%20%26%26%20status%20%3D%20error%20%26%26%20span%3Aname%3D%22add-checkout%22%7D%20%7C%20count_over_time()%20by%20(resource.service.name)&start=1739873095&end=1739874895&step=1800s
```

I think we have a similar issue in the app filter, but there the quotes are added for non-string values. Happy to park this and try to tackle separately.

### Not possible to type regex

When choosing a regex match, I can only select from the drop down, but not enter a regex.

Screenshot 2025-02-18 at 11 39 44 AM

joey-grafana:
@alexbikfalvi I've pushed a fix for the filter query.

As for the regex part, this is something we should discuss and that can be done in another PR.

alexbikfalvi:
@joey-grafana can be filter query be fixed? I understood that it was hard without knowing the attribute type - so it will work for some attributes, but not others.

On regex - now users can type regular expressions. If so, we need to follow up on how to address it, is that right?

joey-grafana:
> @joey-grafana can be filter query be fixed? I understood that it was hard without knowing the attribute type - so it will work for some attributes, but not others.

@alexbikfalvi the filters should be fixed now. 

> On regex - now users can type regular expressions. If so, we need to follow up on how to address it, is that right?

We don't allow this right now but I think we can follow up and address it in another PR. It's more of a mini-feature request.

PR #342, opened 2025-02-20
Select custom columns in trace list

joey-grafana:
Allows users to add custom columns to their trace list.

This is useful for users where the current columns do not offer enough context into their data and there is a particular data point they are searching for.

Screenshot 2025-02-20 at 09 44 53

### DIFF
--- a/src/components/Explore/TracesByService/Tabs/Spans/SpanListColumnsSelector.test.tsx
+++ b/src/components/Explore/TracesByService/Tabs/Spans/SpanListColumnsSelector.test.tsx
@@ -1,0 +1,102 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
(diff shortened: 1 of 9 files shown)


nadinevehling:
As discussed in the recent meeting @joey-grafana 
- We add the input option right-aligned because the columns will be added to the right side.
- Text label: "Add additional columns", Select input text: " *columns icon* Select attribute"

@alexbikfalvi To improve horizontal space and reuse a Grafana pattern, I'd like to explore how we can use the pattern OnCall has for columns in a next iteration (after GA?).

nadinevehling (changes requested):
As discussed in the recent meeting @joey-grafana:
- We add the input option right-aligned because the columns will be added to the right side.
- Text label: "Add additional columns", Select input text: " *columns-icon* Select attribute"

@alexbikfalvi To improve horizontal space and reuse a Grafana pattern, I'd like to explore how we can use the pattern OnCall has for columns in a next iteration (after GA?).

PR #358, opened 2025-03-12
Set Explore route as default view

joey-grafana:


ifrost (approved):

package.json
```diff
@@ -91,5 +91,8 @@
     "tslib": "^2.8.1",
     "zod": "^3.23.8"
   },
+  "resolutions": {
+    "nwsapi": "2.2.16"
```
ifrost, line 95: Without it tests are passing due to https://github.com/dperini/nwsapi/issues/143

PR #363, opened 2025-03-12
Wrap routes in PluginPage

### LINKED ISSUE #116: nit: title starts with `-`
image

adrapereira:
Fixes #116 

image


ifrost:
We need to find a solution that would work with all apps and implement it everywhere. cc @nadinevehling

PR #367, opened 2025-03-13
Add support for bookmarks

joey-grafana:
Adds bookmarks to the homepage. 

Bookmarks allow users to save the following info;

- ACTION_VIEW: `comparison`
- PRIMARY_SIGNAL: `server_spans`
- BOOKMARK_DATA_SOURCE: `EBorgLFZ`
- BOOKMARK_FILTERS: `filter1=value1&&filter2=value2`
- BOOKMARK_GROUPBY: `http.status_code`
- BOOKMARK_METRIC: `duration`

Upon clicking a bookmark, users are taken to the second page with all of the bookmark info filled in making it much easier to get from to the point at which they are seeing the data they are interested in (as requested by multiple users). We used a similar card layout to Metrics Drilldown.

Screenshot 2025-03-13 at 08 55 34

There is a filled in starred icon to represent when the current page is bookmarked (same as Metrics Drilldown).

Screenshot 2025-03-13 at 08 57 49


joey-grafana:
Updates from today include bookmarking the whole url and moving the bookmark icon to the top of the page - as it effectively bookmarks the whole page now as well as many more tests.

Screenshot 2025-03-14 at 14 37 55

PR #371, opened 2025-03-18
Header redesign

adrapereira:


alexbikfalvi:
Also, I was checking the issue with the value types, and now this happens - same local dev environment + data source but before and after:

Commit `abe282a`:

Screenshot 2025-03-19 at 10 18 46 AM

Branch: `main`:

Screenshot 2025-03-19 at 10 20 02 AM

Issue #377, opened 2025-03-24
Add streaming indicator to the RED metrics panels


alexbikfalvi:
### Why is this needed?

Currently, when streaming is enabled there is no streaming indicator on the RED metrics panels indicating that the data is loading.

Thus, users cannot tell if the data is complete or it's still in progress of being loaded. Adding a streaming indicator will give a clear visual cue, and will make it consistent with the other metrics panels.

![Image](https://github.com/user-attachments/assets/d1b87053-03ca-4870-bb70-606edc889073)

### What should we add

Add a streaming indicator as for the other metrics panels.

PR #382, opened 2025-03-25
Show streaming indicator for RED panels

joey-grafana:
Fixes https://github.com/grafana/traces-drilldown/issues/377

Uploading Screen Recording 2025-03-25 at 11.34.22.mov…


src/components/Explore/StreamingIndicator.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { css } from '@emotion/css';
+
+interface StreamingIndicatorProps {
+  isStreaming: boolean;
+  iconSize?: number;
+}
+
+export const StreamingIndicator = ({ 
+  isStreaming, 
+  iconSize = 14,
```
ifrost, line 13: It seems we always pass the size. Do we need the default?

ifrost (approved):

PR #384, opened 2025-03-25
Make exemplars more distinguishable

joey-grafana:
Switches to using a line draw style, with a fill opacity of 15% which makes it much easier to see the exemplars.

Screenshot 2025-03-25 at 15 24 20


nadinevehling:
I don't have a strong opinion here, and I leave the decision up to you. If this is connected to the description of the y-axis being "rate" now, then this is clearer. If we can display bars with a good y-axis description, I support this, too.

joey-grafana:
Screenshot 2025-03-31 at 10 14 58

PR #386, opened 2025-03-25
Gracefully handle metrics query error when max range is exceeded

### LINKED ISSUE #378: Graceful handling of exceeding the time range limit
### Why is this needed?

By default, TraceQL metrics queries only support up to 24h time ranges. While some tenants have requested exceptions to this (and may have up to 48h), for most users it's relatively easy to exceed this range, especially when using Grafana's time range picker, with which you can easily select a longer time range.

Image

However, when users select a longer range, the page fails with an error, which often is too generic as shown below.

Image

Sometimes, the error includes more details leaving some clues on what is wrong

Image

Nevertheless, users must be familiar with the limitations of Tempo in order to interpret this error as exceeding the backend limit.

### What should be added

Catch and handle the error more gracefully, with a custom error message telling users that they are reaching the query time range limit. We can emphasize that the range limit is a sliding window: they could still choose a similar range in any part of their retention period (usually 30 days).

Future iterations may include a smarter range picker that limits or encourages ranges only up to the permitted limit. However, we should still handle the error case gracefully, in case it's passed as a URL argument.

joey-grafana:
Fixes https://github.com/grafana/traces-drilldown/issues/378

### DIFF
--- a/src/components/Explore/ByFrameRepeater.tsx
+++ b/src/components/Explore/ByFrameRepeater.tsx
@@ -72,7 +72,7 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
                 new SceneCSSGridLayout({
                   children: [
                     new ErrorStateScene({
-                      message: data.data.errors?.[0]?.message ?? 'An error occurred in the query',
+                      message: (data.data.errors?.[0]?.message || data.data.error?.message) ?? 'An error occurred in the query',
                     }),
                   ],
                 }),


src/components/Explore/ByFrameRepeater.tsx
```diff
@@ -72,7 +72,7 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
                 new SceneCSSGridLayout({
                   children: [
                     new ErrorStateScene({
-                      message: data.data.errors?.[0]?.message ?? 'An error occurred in the query',
+                      message: (data.data.errors?.[0]?.message || data.data.error?.message) ?? 'An error occurred in the query',
```
ifrost, line 75: [This field is deprecated](https://github.com/grafana/grafana/blob/v11.6.0/packages/grafana-data/src/types/panel.ts#L65), we should probably not rely on it.

ifrost (changes requested):

Issue #368, opened 2025-03-17
Error rate bar coloring is mistakenly colored green if query expressions contains "ok" - as in `loki-querier`


svennergr:
The field configuration in https://github.com/grafana/traces-drilldown/blob/abe282a66d86e19c077c5f5f0cec38fcf0ed895b/src/components/Explore/panels/barsPanel.ts#L14-L25 colors series conditionally on their field name. The field name is constructed with the query expression. E.g. querying for `{resource.service.name="foo"}` will also result in a field name `{resource.service.name="foo"}`.

Combined with the [field configuration](https://github.com/grafana/traces-drilldown/blob/abe282a66d86e19c077c5f5f0cec38fcf0ed895b/src/components/Explore/panels/barsPanel.ts#L14-L25) this means, that every "service name", that contains `ok` will be colored green. Just like this example where the service name was `loki-querier`:

![Image](https://github.com/user-attachments/assets/89fe05f9-fe77-40fc-81cc-471cac31392e)

PR #387, opened 2025-03-26
Fix regex matcher for field names

joey-grafana:
Fixes issue where fields with names like `"{resource.service.name=\"loki-ingester\", status=\"error\"}"` were getting matched as ok.

Now we ensure either the field name has `status="error"` or the entire field name is just set to `error`.

Fixes https://github.com/grafana/traces-drilldown/issues/368


ifrost (approved):
Works great!

Issue #385, opened 2025-03-25
Exemplars: Clarify why they are shown in different styles


joey-grafana:
Crosses and squared bullets (diamonds).

![Image](https://github.com/user-attachments/assets/6d2c3eb7-f72b-4a06-b402-84f544b309a5)

PR #388, opened 2025-03-26
Fix for exemplars with two different styles

joey-grafana:
We were seeing exemplars with two different styles as this is actually how exemplars work when they are receiving two different sets of labels. 

In our case, our metrics rate query was also returning series with error labels and so the exemplar code was showing the diamond for the one exemplar field and the cross for the other exemplar field.

I've updated the query so that we don't return error series when we run a rate query.

Bonus: we will no longer see red in our rate query graphs! 🥳 

Fixes https://github.com/grafana/traces-drilldown/issues/385

### DIFF
--- a/src/components/Explore/queries/generateMetricsQuery.ts
+++ b/src/components/Explore/queries/generateMetricsQuery.ts
@@ -11,7 +11,9 @@ export function generateMetricsQuery({ metric, groupByKey, extraFilters, groupBy
   // Generate span set filters
   let filters = `${VAR_FILTERS_EXPR}`;
 
-  if (metric === 'errors') {
+  if (metric === 'rate') {
+    filters += ' && status!=error';
+  } else if (metric === 'errors') {
     filters += ' && status=error';
   }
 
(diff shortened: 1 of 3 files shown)


src/components/Explore/queries/generateMetricsQuery.ts
```diff
@@ -11,7 +11,9 @@ export function generateMetricsQuery({ metric, groupByKey, extraFilters, groupBy
   // Generate span set filters
   let filters = `${VAR_FILTERS_EXPR}`;
 
-  if (metric === 'errors') {
+  if (metric === 'rate') {
+    filters += ' && status!=error';
```
ifrost, line 15: Hmm... 🤔  Does it mean we drop errors altogether in rate metric? Wouldn't it be confusing (e.g. if there are only errors the rate would be 0)?

| before | after |
|---|---|
| Screenshot 2025-03-26 at 13 24 27 | Screenshot 2025-03-26 at 13 24 32 |

joey-grafana:
Thanks for the reviews! 

I can make those changes. However, I'm not sure if it was mentioned before that for a few of our user interviews there was confusion as to why we are actually showing errors for the rate at all, as it overlaps with the errors card. We understand why and it makes sense from our perspective as developers but I think users often understand this as the rate of errors or the rate of non errors cc @nadinevehling

nadinevehling:
I don't have data on this. Overall, I think errors and duration are a clear pattern. For rate, there is sometimes more confusion around this for some engineers. That's why we have the "Need help" toggle component.
What I can say is that if we keep the errors in rate, I wouldn't use a different pattern for exemplars but keep one.

PR #393, opened 2025-04-01
Only show valid exemplars

### LINKED ISSUE #390: Exemplars shown at or beyond the edge of the chart
### What happened?

Often, the RED metrics visualizations show exemplars at the right hand size edge, before showing and chart data. Sometimes the exemplars are drawn even outside of the panel.

When hovering over them, a pop over shows valid exemplar information.

Image

joey-grafana:
Ensure we do not try to render exemplars which have invalid data.

A exemplar is valid if:
1. The traceId is not null/undefined
2. The value is not null/undefined
3. The value is not 0

The data is coming from the backend so it may also be worth improving the API response.

Fixes https://github.com/grafana/traces-drilldown/issues/390

### DIFF
--- a/src/utils/exemplars.test.ts
+++ b/src/utils/exemplars.test.ts
@@ -12,8 +12,8 @@ describe('exemplarsTransformations', () => {
   };
 
   const createExemplarFrame = (
-    traceIds: (string | null | undefined)[], 
-    values: (number | null | undefined)[]
+    traceIds: Array<string | null | undefined>, 
+    values: Array<number | null | undefined>
   ): DataFrame => {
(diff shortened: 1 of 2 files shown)


adrapereira (changes requested):
Can we please make this change in the datasource? 🙏 
Invalid exemplars should never be returned by the datasource backend in Grafana.

joey-grafana:
Yes, that is a solid suggestion! I'll take a look.

joey-grafana:
Closing in favor of https://github.com/grafana/grafana/pull/103298

Issue #244, opened 2024-09-26
Structural Streaming causes "flashing" and resets UI


joe-elliott:
When using the structural tab with streaming over a large dataset, the structural tab "flickers" as it constantly rerenders.

This also causes the "View linked Span" drop down to disappear which makes the UI difficult to interact with.

PR #394, opened 2025-04-02
Fix structure tab flickering

joey-grafana:
This PR makes two improvements that now prevent the structure tab from flickering when receiving data

- skips duplicate frames received within 100ms of each other,
- prevents multiple UI updates in the same 500ms

Fixes https://github.com/grafana/traces-drilldown/issues/244


src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
```diff
@@ -55,34 +60,88 @@ export class StructureTabScene extends SceneObjectBase<ServicesTabSceneState> {
...
-            tree: merged,
-            panel: new SceneFlexLayout({
-              height: '100%',
-              wrap: 'wrap',
-              children: this.getPanels(merged),
-            }),
-          });
+          const now = Date.now();
+          // Prevent duplicate processing of the same data
+          // Skip if we processed this exact data less than 100ms ago
```
ifrost, line 92: > Skip if we processed this exact data less than 100ms ago

If it's exactly the same data do we need to do any processing at all or could be skip it altogether? 🤔

src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
```diff
@@ -55,34 +60,88 @@ export class StructureTabScene extends SceneObjectBase<ServicesTabSceneState> {
...
+  private _onDeactivate() {
+    if (this.updateTimeoutId) {
+      clearTimeout(this.updateTimeoutId);
+      this.updateTimeoutId = null;
+    }
+    this.lastProcessedData = null;
   }
 
   public _onActivate() {
     this.state.$data?.subscribeToState((state) => {
```
ifrost, line 78: BTW. is this subscription cleared? To get it automatically removed by scenes I think it should be wrapped in `this._subs.add(...)`

src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
```diff
@@ -55,34 +60,88 @@ export class StructureTabScene extends SceneObjectBase<ServicesTabSceneState> {
...
+                this.updateTimeoutId = null;
+              }, STREAMING_DEBOUNCE_TIME_MS);
+              
+              return;
+            }
+            
+            // For done state, update immediately
+            this.updatePanels(merged);
+          } catch (error) {
+            console.error('Error processing trace data:', error);
```
ifrost, line 124: Should we show the error to the user or we intentionally skip it silently? Would it make sense to push it at least to faro?

src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
joey-grafana: Nice catch, added! 👍

src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
joey-grafana: This is returned immediately if it's duplicate data though 🤔

src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
joey-grafana: I've removed for now as it's not necessary for the PR.

Issue #379, opened 2025-03-24
Ensure Y-axis label matches the data for RED metrics


alexbikfalvi:
### Why is this needed?

The Y-axis label for rate and errors it's not easy to deduce just by using the app what the numbers represent:

* For rate, it's unclear whether the "rate" numbers represent spans/s, spans/min, etc.
* For errors, it's unclear whether the numbers represent "error count" or error rate (errors/s, errors/min, etc.)

We should make it easier for users to understand what the numbers represent.

### What should we change

Throughput and error metrics are usually shown as a rate (e.g., requests per second, errors per second or as a percentage) because it provides a normalized view that is easier to compare across different time intervals or varying traffic volumes. This normalization helps operators quickly assess system performance and health, regardless of the underlying traffic level. Read more: https://sre.google/sre-book/service-level-objectives/

Use rate for both throughput and errors, indicate the correct the correct unit such that users can easily interpret the numbers (such as traces/s, spans/s, errors/s).

PR #401, opened 2025-04-04
Ensure Y-axis label matches the data for RED metrics

joey-grafana:
Fixes https://github.com/grafana/traces-drilldown/issues/379

Tried first putting the error text on the y-axis as well but the text is too long in the MiniRedPanel. I don't think the errors text is required anyway as it's already in the title and is quite obvious from the red color.

Screenshot 2025-04-04 at 12 40 39

Looks better with just `span/s`.

Screenshot 2025-04-04 at 12 40 46

### DIFF
--- a/src/components/Explore/TracesByService/MiniREDPanel.tsx
+++ b/src/components/Explore/TracesByService/MiniREDPanel.tsx
@@ -116,7 +116,7 @@ export class MiniREDPanel extends SceneObjectBase<MiniREDPanelState> {
     if (metric === 'rate') {
       panel.setCustomFieldConfig('axisLabel', 'span/s');
     } else if (metric === 'errors') {
-      panel.setCustomFieldConfig('axisLabel', 'span/s').setColor({
+      panel.setTitle('Errors rate').setCustomFieldConfig('axisLabel', 'span/s').setColor({
         fixedColor: 'semi-dark-red',
         mode: 'fixed',
       });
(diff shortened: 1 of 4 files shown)


src/components/Explore/TracesByService/MiniREDPanel.tsx
```diff
@@ -113,13 +113,13 @@ export class MiniREDPanel extends SceneObjectBase<MiniREDPanelState> {
 
   private getRateOrErrorPanel(metric: MetricFunction) {
     const panel = barsPanelConfig().setHoverHeader(true).setDisplayMode('transparent');
-    if (metric === 'errors') {
-      panel.setTitle('Errors rate').setCustomFieldConfig('axisLabel', 'Errors').setColor({
+    if (metric === 'rate') {
+      panel.setCustomFieldConfig('axisLabel', 'span/s');
```
ifrost, line 117: Is it always `span/s` or can be `span/min`, `span/h`, etc. depending on final resolution?

joey-grafana: Hmm good question. I'm not actually [sure](https://github.com/grafana/traces-drilldown/blob/main/src/components/Explore/queries/generateMetricsQuery.ts#L54). @adrapereira do you have any insight here?

ifrost (approved):

PR #404, opened 2025-04-09
Explore: Add "Go Queryless" hook

### LINKED ISSUE #341: Button to quickly navigate from Explore to Drilldown
### Why is this needed?

We want the users to easily discover the queryless traces experience enabled by Traces Drilldown.

### What do we need to add?

Similarly to Loki and Pyroscope, add a "Go queryless" button in Explore when the data source is Tempo.

Image

Image

ifrost:
Fixes #341

It creates the same link as the extension we currently have for dashboards panels.

We don't have a good icon in core Grafana to use :( but we can add something and update it later.

https://github.com/user-attachments/assets/46fec9f7-541e-4e39-bce5-29b7970afba8

### DIFF
--- a/src/utils/links.ts
+++ b/src/utils/links.ts
@@ -1,12 +1,12 @@
 import {
   PluginExtensionAddedLinkConfig,
   PluginExtensionPanelContext,
-  PluginExtensionPoints,
+  PluginExtensionPoints, RawTimeRange,
   toURLRange
 } from '@grafana/data';
 
-import { DataSourceRef } from '@grafana/schema';
-import { EXPLORATIONS_ROUTE, VAR_DATASOURCE, VAR_FILTERS, VAR_METRIC } from './shared';
+import {DataQuery, DataSourceRef} from '@grafana/schema';
+import {EXPLORATIONS_ROUTE, VAR_DATASOURCE, VAR_FILTERS, VAR_METRIC} from './shared';
 
 type TempoQuery = {
   filters?: TraceqlFilter[];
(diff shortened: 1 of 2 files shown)


src/utils/links.ts
```diff
@@ -20,20 +20,37 @@ export interface TraceqlFilter {
...
+    description: 'Try our new queryless experience for traces',
+    path: createAppUrl(),
+    configure: (context?: PluginExtensionExploreContext) => contextToLink(context),
+  }
 ];
 
-export function contextToLink<T extends PluginExtensionPanelContext>(context?: T) {
+/**
+ * Props shared between ...
+ */
```
joey-grafana, line 47: Do we want to update this?

joey-grafana (approved):
Thanks for taking this one 👍 

One thing I noticed, `name` becomes `span.name` in the app.

Screenshot 2025-04-09 at 15 21 28
Screenshot 2025-04-09 at 15 21 35
Screenshot 2025-04-09 at 15 24 41
Screenshot 2025-04-09 at 15 24 56

src/utils/links.ts
ifrost: Thanks 🤦 I've made it more explicit in 85ba35f1f3baeef5577bc2487ace4b901fe71bb8

ifrost:
> One thing I noticed, `name` becomes `span.name` in the app.

Good catch! It's an interesting one. We already have this problem when going from a dashboard panel. It seems like we always use `scope.tag` format but for intrinsics it should be `scope:tag`. We don't have access to the list created dynamically so ideally we'd need this to be passed inside the query (though that would require more changes in grafana/grafana). Maybe we can have a list of v2 intrinsic fields like in https://github.com/grafana/traces-drilldown/pull/404/commits/2ccbed2ec4aeb0028a30b1573edf70138cc2a7d3?

Also, do you know why we skip `status` field with `.filter((filter) => filter.tag !== 'status')` ?

joey-grafana:
Yeah I think we will need another list of intrinsics in the app. Technically the list has what we need and acts as a fallback, but we also query for the intrinsics in core Grafana so we ensure we use the most up to date list of intrinsics, although that may be overkill here.

IIRC the `status` field is filtered because the `statusFilter` is added [separately](https://github.com/grafana/traces-drilldown/blob/main/src/utils/links.ts#L58).

PR #407, opened 2025-04-10
New exposed component to embed the trace exploration scene

adrapereira:
This PR is a POC of how we can embed Traces Drilldown in other apps. It exposes a component that other apps can use and customise to fit their use case. The exposed component currently allows for the following props:
- **time range**
- **onTimeRangeChange** callback function which is called whenever the time range in Traces Drilldown changes
- **initialFilters**: set of filters used to initialise the ad-hoc filters that will become read only, however users are still be able to add new filters
- **initialActionView**: the default value for the action view, aka the tabs, for example `breakdown`
- **allowedActionViews**: the list of action views (tabs) that should be rendered, for example `['breakdown', 'traceList']`. If the list only contains one element the app will render the content for that tab but won't render the tabs selector. 
- **initialGroupBy**: the attribute that should be used for the group by selector, for example `name` or `resource.namespace`

Also improved the group by selector to be smarter when changing scopes and when an attribute is added to the filters. The component will automatically select the next candidate to avoid the problem where the breakdown tab would be useless after adding a filter for the currently selected attribute. 

Additionally refactored the code to open traces and removed the dependency on the `locationService`.

https://github.com/user-attachments/assets/608cb3b0-261b-451b-8ae4-01cc7f135e75

### DIFF
--- a/src/module.tsx
+++ b/src/module.tsx
@@ -2,8 +2,6 @@ import { lazy } from 'react';
 import { AppPlugin } from '@grafana/data';
 
 // @ts-ignore new API that is not yet in stable release
-import { sidecarServiceSingleton_EXPERIMENTAL } from '@grafana/runtime';
-import pluginJson from './plugin.json';
 import { EmbeddedTraceExplorationState, OpenInExploreTracesButtonProps } from 'exposedComponents/types';
 import { SuspendedEmbeddedTraceExploration, SuspendedOpenInExploreTracesButton } from 'exposedComponents';
 import { linkConfigs } from 'utils/links';
@@ -18,15 +16,6 @@ export const plugin = new AppPlugin<{}>()
     icon: 'cog',
     body: AppConfig,
     id: 'configuration',
-  })
-  .addLink({
-    title: 'Traces Drilldown',
-    description: 'Open in Traces Drilldown',
-    icon: 'align-left',
-    targets: 'grafana-lokiexplore-app/toolbar-open-related/v1',
-    onClick: () => {
-      sidecarServiceSingleton_EXPERIMENTAL?.openAppV3({ pluginId: pluginJson.id, path: '/explore' });
-    },
   })
   .exposeComponent({
     id: 'grafana-exploretraces-app/open-in-explore-traces-button/v1',
--- a/src/pages/Explore/TraceExplorationPage.tsx
+++ b/src/pages/Explore/TraceExplorationPage.tsx
@@ -1,25 +1,14 @@
 import React, { useEffect, useState } from 'react';
-import z from 'zod';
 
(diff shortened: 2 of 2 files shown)


src/pages/Explore/TraceExploration.tsx
```diff
@@ -371,40 +330,168 @@ export class TraceExplorationScene extends SceneObjectBase {
...
+        <Stack gap={1} alignItems={'center'}>
+          <LinkButton
+            href={getUrlForExploration(traceExploration)}
+            variant="secondary"
+            icon="arrow-right"
+            onClick={() => {
+              if (returnToPreviousSource) {
+                setReturnToPrevious(returnToPreviousSource);
+              }
+              reportAppInteraction(USER_EVENTS_PAGES.home, USER_EVENTS_ACTIONS.home.explore_traces_clicked);
```
joey-grafana, line 362: This should be updated?

joey-grafana (changes requested):
Nice work 👍 I really like the subtle animation when opening opening the trace view drawer - although I know we are prob going to move away from a drawer on a drawer 😆 

I can modify/remove the filters, it was my understanding that this is not supposed to happen.
Screenshot 2025-04-25 at 11 49 17

There are a lot of inconsistencies with the Logs tab. For example, they show the selected data source (disabled), which is helpful so users know where their data is coming from. Logs allow you to refresh your query which feels like something the user would want to do. Also, the position of Open in Logs Drilldown is different, and does not have an icon. 
Screenshot 2025-04-25 at 11 54 52
Screenshot 2025-04-25 at 11 55 13

Why/where are we grouping by `container.id`? What if it or no data exists?
Screenshot 2025-04-25 at 12 02 54

I changed the time range from last 30 mins to last 5 mins and caused some kind of exception which removed all the tabs!
Screenshot 2025-04-25 at 12 08 53
Screenshot 2025-04-25 at 12 10 57




Why is the `Open in Traces Drilldown` button not opening up a new tab? It feels like something that should open in a new tab so just wondering if this has been discussed with UX? The links in the trace view + read documentation in the tooltip open in a new tab.


Autocomplete / tag suggestions are not working in the filters bar.

src/pages/Explore/TraceExploration.tsx
joey-grafana: The `reportAppInteraction ` that is.

joey-grafana (approved):
Code changes / design LGTM as we've decided to hold off on making some of the other improvements until later 👍 

Note that the data we've set up locally is terrible (currently discussing) and makes this pretty hard to test with proper traces to play around so this review is mostly based on memory from the previous review and what has changed since! Also adding filters in a WIP and will be added in another PR.

Not sure about the compatibility check though!

PR #410, opened 2025-04-10
Use events to open traces

### LINKED ISSUE #406: Query error when viewing all spans and opening a trace
### What's happening?

When selecting "all spans" and opening a trace, the query fails:

https://github.com/user-attachments/assets/bf3e22f2-ca0d-49bd-89e4-968dc6e6fb31

https://github.com/user-attachments/assets/177fb1fd-9f95-43fd-bfb1-932b6510dbbd

adrapereira:
Fixes issue where the primary signal would become unselected when opening a trace.

Fixes #406


joey-grafana (changes requested):
Can you create a shared `openTrace` method please? I would prefer for debugging / maintainability purposes if all `EventTraceOpened` events are routed through the same method.

joey-grafana (approved):

PR #427, opened 2025-04-15
Order radio options and dropdown by frequently used

### LINKED ISSUE #302: Make the default GroupBy attributes customizable
### User problem

In Explore Traces, traces can be grouped using the `Group By`. At this moment for us by default `Group By` shows two default label `cluster` and `environment`. However, users may have a different attribute `resource.some_env` to actually representing the environment for the traces and there is no `environment` for the traces.

As the `environment` shows as default in `Group By`, our users try to group the traces based on `environment` and get an empty result which is confusing for them. In my opinion, having some default is completely fine but having an irrelevant (`environment`) default value is confusing for general users.

Users think it would be less confusing for users if the `Group By` section is grayed out with an example option instead of a button (cluster, environment) which convinces them that these are valid values for our data. Otherwise if they could have only default values that are relevant to their traces that would also help.

### What is the problem the customer is trying to solve

Improve the defaults to make it easier for their users as they don't use any of the values we include in our defaults.

joey-grafana:
Orders the `GroupBySelector` radio options and the dropdown next to it by attribute clicks. The attributes with the most clicks will move towards the far left for the radio options and the top of the frequently used option group in the dropdown.

Fixes https://github.com/grafana/traces-drilldown/issues/302

https://github.com/user-attachments/assets/2c905952-6700-44c6-8e82-ed5ffb93f62b

### DIFF
--- a/src/components/Explore/GroupBySelector.tsx
+++ b/src/components/Explore/GroupBySelector.tsx
@@ -30,6 +30,8 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
 
   const [selectQuery, setSelectQuery] = useState<string>('');
   const [clickCounts, setClickCounts] = useState<Record<string, number>>({});
+  // Store initial click counts separately for sorting to prevent reordering on clicks
+  const [initialClickCounts, setInitialClickCounts] = useState<Record<string, number>>({});
 
   const [availableWidth, setAvailableWidth] = useState<number>(0);
   const controlsContainer = useRef<HTMLDivElement>(null);
@@ -45,6 +47,7 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
         if (storedCounts) {
           const counts = JSON.parse(storedCounts);
           setClickCounts(counts);
+          setInitialClickCounts(counts);
         }
       } catch (e) {
         console.error('Failed to load group by click counts:', e);
@@ -115,7 +118,7 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
       .filter(option => 
         isOptionValid(option) && 
         !radioAttributes.includes(option) && 
-        (clickCounts[option] || 0) > 0
+        (initialClickCounts[option] || 0) > 0
       );
 
     const allOptions = [...validRadioOptions, ...validOtherOptions];
@@ -124,7 +127,7 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
       allOptions,
       isOptionValid
     };
-  }, [options, radioAttributes, filters, metricValue, clickCounts]);
+  }, [options, radioAttributes, filters, metricValue, initialClickCounts]);
 
   const radioOptions = useMemo(() => {
     let radioOptionsWidth = 0;
@@ -135,7 +138,7 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
         label: attribute.replace(SPAN_ATTR, '').replace(RESOURCE_ATTR, ''),
         text: attribute,
         value: attribute,
-        clickCount: clickCounts[attribute] || 0
+        clickCount: initialClickCounts[attribute] || 0
       }))
       .sort((a, b) => b.clickCount - a.clickCount)
       .filter((option) => {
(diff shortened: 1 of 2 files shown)


joey-grafana:
@ifrost please take a look when you can :) Would like to get this PR merged in next few days

src/components/Explore/GroupBySelector.tsx
```diff
@@ -35,6 +38,38 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
...
+    
+    loadClickCounts();
+  }, []);
+
+  const trackOptionClick = async (option: string) => {
+    try {
+      const updatedCounts = { ...clickCounts };
+      updatedCounts[option] = (updatedCounts[option] || 0) + 1;
+      setClickCounts(updatedCounts);
+      await storage.setItem(GROUP_BY_CLICK_COUNTS_STORAGE_KEY, JSON.stringify(updatedCounts));
```
ifrost, line 62: The issue description says

> This would work per user and browser, without any configuration

but we save it to user storage in the backend. Is it what we want in this case or it could be an overkill for the storage?

src/components/Explore/GroupBySelector.tsx
```diff
@@ -45,37 +80,64 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
...
+    const { allOptions } = filterValidOptions;
+    
+    return allOptions
       .map((attribute) => ({
         label: attribute.replace(SPAN_ATTR, '').replace(RESOURCE_ATTR, ''),
         text: attribute,
         value: attribute,
+        clickCount: clickCounts[attribute] || 0
       }))
+      .sort((a, b) => b.clickCount - a.clickCount)
```
ifrost, line 140: It works as described, though personally I find it a bit unclear what the behavior is (not knowing it upfront). The items when clicked start to jump around and it might be hard to figure out what is going on based on UI. A few things that would help me:

- A tooltip describing how options are sorted
- Sorting only on load (e.g. after reload) not on each click to reduce items jumping back and forth

Alternatively we could allow manually "pinning" items, or dismissing items instead of promoting them but that would require a lot of UI changes.

src/components/Explore/GroupBySelector.tsx
```diff
@@ -45,37 +80,64 @@ export function GroupBySelector({ options, radioAttributes, value, onChange, sho
...
+  const radioOptions = useMemo(() => {
+    let radioOptionsWidth = 0;
+    const { allOptions } = filterValidOptions;
+    
+    return allOptions
       .map((attribute) => ({
         label: attribute.replace(SPAN_ATTR, '').replace(RESOURCE_ATTR, ''),
         text: attribute,
         value: attribute,
+        clickCount: clickCounts[attribute] || 0
```
ifrost, line 138: There might be two potential issues with counting total clicks per label

1) The value is absolute per stack, not scoped to the selected data source, but each data source may have different set of labels to be prioritized. 
2) Using total values will create high values after some time making it hard for new labels to become favorites

src/components/Explore/GroupBySelector.tsx
joey-grafana: I think that is not a requirement that it's per browser, that was just a description based on the conversation we had at the time. Userstorage is the preferred Grafana way to store user preferences etc so not sure how you would measure if this is overkill.

src/components/Explore/GroupBySelector.tsx
joey-grafana: This is a good tip, will look into it.

src/components/Explore/GroupBySelector.tsx
ifrost: 1. 👍  
2. Like a button in the UI to reset? That may clatter UI 🤔  I'll leave it up to @alexbikfalvi and @nadinevehling. I think pinning/unpinning could work a bit better but we can go with what we have in here and get feedback 👍

ifrost:
FYI: similar sorting based on recent usage in Metrics Drilldown (https://github.com/grafana/metrics-drilldown/pull/282/)

Issue #411, opened 2025-04-10
Duration empty state


joey-grafana:
Duration does not have an empty state on the slow traces tab

PR #430, opened 2025-04-16
Add fix to show empty state in the trace list

joey-grafana:
Fixes https://github.com/grafana/traces-drilldown/issues/411


joey-grafana:
No problem. Checked and it does cause another render but not an infinite loop. 

For good measure, I put back in the check and added a new one to check if the panel already exists before returning.

Issue #415, opened 2025-04-10
Comparison tab issue with highest difference


joey-grafana:
Sometimes the values are “infinity” or larger than 100%, which shouldn’t be possible. Only seen locally and not sure how to reproduce. [André]. 

To Reproduce
..Duration metric in ops or if long standing streaming (more than 1 min), once it moves over to the next minute, it will re-trigger a new query. 

Possible Fix
Maybe lock time range.

PR #435, opened 2025-04-18
Fix to normalize comparison data when total fields are missing

joey-grafana:
Fixes https://github.com/grafana/traces-drilldown/issues/415


ifrost (approved):
I couldn't reproduce it locally but looks legit 👍  I've tested it with following unit tests: https://github.com/grafana/traces-drilldown/pull/436 (feel free to merge it into your branch).

PR #440, opened 2025-04-23
Show red for errors in graph

joey-grafana:
In order to get the colors in the graph, we need to group by status, so that the status is returned in the labels.


ifrost:
Could you add some description to the PR for the record? Is it adding grouping by status to automatically get correct colors on the graph?

joey-grafana:
> Could you add some description to the PR for the record? Is it adding grouping by status to automatically get correct colors on the graph?

Done

PR #451, opened 2025-04-29
[DOC] Add note for app with G12

knylander-grafana:
Add note about Traces Drilldown possibly being included in a G12 instatllation.

### DIFF
--- a/docs/sources/access/index.md
+++ b/docs/sources/access/index.md
@@ -43,6 +43,11 @@ Next, [access Traces Drilldown](#access-traces-drilldown).
 
 ### Install the Traces Drilldown plugin
 
+{{< admonition type="note" >}}
+If you are using Grafana v12 or later, Traces Drilldown may be pre-installed.
+To check, refer to [Access Traces Drilldown](#access-traces-drilldown).
+{{</ admonition >}}
+
 Traces Drilldown is distributed as a Grafana Plugin.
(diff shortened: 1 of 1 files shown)


joey-grafana:
@alexbikfalvi @knylander-grafana not sure if you wanted to update the text here or if it's necessary. But let's please get this merged today :)

docs/sources/access/index.md
```diff
@@ -43,6 +43,11 @@ Next, [access Traces Drilldown](#access-traces-drilldown).
 
 ### Install the Traces Drilldown plugin
 
+{{< admonition type="note" >}}
+If you are using Grafana v12 or later, Traces Drilldown may be pre-installed.
```
alexbikfalvi, line 47: @adrapereira confirmed that Traces Drilldown is available by default after upgrading to Grafana 12. So we can rewrite this as something like
```suggestion
If you are using Grafana v12 or later, Traces Drilldown is already installed, go to [Access Traces Drilldown](#access-traces-drilldown).
```

Also we can consider that at some point, this will be the default. So this section would become: only needed if you're used Grafana 11

alexbikfalvi: @joey-grafana @knylander-grafana left a comment - not sure what's the best way to write it, we can leave the admonition for now, but just say this installation is not needed for versions >= G12.

joey-grafana (approved):

PR #456, opened 2025-05-01
Fix Zizmor persist credentials issues

joey-grafana:


ifrost (approved):

joey-grafana:
@adrapereira We will see this compatibility check fail on every PR now until we remove that code. It's for 11.x users to retain the functionality for plugin extensions and 12.0 is just going out the door with the new API so how long do you suggest we keep that for?

PR #459, opened 2025-05-07
Fix exposing types

adrapereira:
Refactors the type of TraceExplorationState so we can expose them in `plugin-extension-types` and use them in other plugins. 

Also adds `levignore` to ignore the compatibility errors with Grafana 12.


joey-grafana (approved):
Didn't checkout and run but it's just exposed types so should be fine!

Issue #463, opened 2025-05-08
Preserve Asserts context in Traces Drilldown


alexbikfalvi:
### Why is this needed

When using Traces Drilldown in Asserts, we want users to preserve their assertions context while drilling down through traces towards finding the root cause of the issue they're investigating.

Having this would easily enable them to access the conditions that triggered the investigation, without leaving Traces Drilldown.

### What do we want to add

In an initial iteration, this could be done as follows:

![Image](https://github.com/user-attachments/assets/0e949bd6-b712-41da-abcd-92313a854e9e)

PR #464, opened 2025-05-09
Preserve asserts context via embedded assertions widget component

joey-grafana:
When using Traces Drilldown in Asserts, we want users to preserve their assertions context while drilling down through traces towards finding the root cause of the issue they're investigating.

This easily enables them to access the conditions that triggered the investigation, without leaving Traces Drilldown.

Fixes https://github.com/grafana/traces-drilldown/issues/463

Screenshot 2025-05-21 at 11 19 24
Screenshot 2025-05-21 at 11 19 34


src/addedComponents/EntityAssertionsWidget/usePluginComponent.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function usePluginComponent<T extends object>(
+  id: string
+): {
+  isLoading: boolean;
+  component: React.ComponentType<T> | null | undefined;
+} {
+  if (realUsePluginComponent) {
+    return realUsePluginComponent<T>(id);
+  } else {
```
ifrost, line 11: Why do we need to wrap it like this? I what case `usePluginComponent` is not defined?

joey-grafana: Yeah we don't need that in fairness - API has been pushed about 9 months ago! I'll remove it.

src/addedComponents/EntityAssertionsWidget/EntityAssertionsWidget.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  return (
+    <EntityAssertionsWidgetExternal
+      size='md'
+      source='Traces Drilldown'
+      query={{
+        start: range.from.valueOf(),
+        end: range.to.valueOf(),
+        entityName: serviceName,
```
ifrost, line 41: Might be another example where we could using OTEL's `Resource` for passing service name like in https://github.com/grafana/grafana/pull/104680#discussion_r2107188381 🤔 . Just food for thought, nothing for this PR.

src/pages/Explore/TraceExploration.tsx
```diff
@@ -307,6 +308,31 @@ export class TraceExplorationScene extends SceneObjectBase {
   };
 }
 
+const useServiceName = (model: SceneObject) => {
```
ifrost, line 311: Are there any plans to use the EntityAssertionsWidget in other places? In theory the logic to get service name could be encapsulated in that component (though it would require passing `model` or making it a scene object).

ifrost (approved):

src/pages/Explore/TraceExploration.tsx
```diff
@@ -414,6 +448,10 @@ const TraceExplorationHeader = ({ controls, model }: TraceExplorationHeaderProps
           )}
         </Stack>
         <div className={styles.controls}>
+          <EntityAssertionsWidget
```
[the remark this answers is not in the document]
joey-grafana: I don't think so tbh.

PR #466, opened 2025-05-12
Embedded mode improvements

adrapereira:
This PR improves URL syncing when the app is running in embedded mode, by wrapping the TraceExploration scene in a `UrlSyncContextProvider` and initialising the state from the URL. This must be enabled by setting `urlSync` to `true` when calling the embedded component.  

Also fixes an issue where there was no primary signal selected #469 

Lastly, when embedded, the drawer that opens with the trace view will be replaced with a modal-like element that will fill all the available space and be rendered on top of the app, with a button to go back.

https://github.com/user-attachments/assets/572e5056-b896-41b0-9f1e-183c64590061

### DIFF
--- a/src/components/App/globalStyles.css
+++ b/src/components/App/globalStyles.css
@@ -1,5 +1,0 @@
-@media screen and (max-width: 1300px) {
-  .rc-drawer-content-wrapper {
-    width: 93vw !important;
-  }
-}
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -3,8 +3,6 @@ import { AppRootProps, PageLayoutType } from '@grafana/data';
 import { AppRoutes } from '../Routes';
 import { PluginPage } from '@grafana/runtime';
 
-import './globalStyles.css';
-
 // This is used to be able to retrieve the root plugin props anywhere inside the app.
 const PluginPropsContext = React.createContext<AppRootProps | null>(null);
 
--- a/src/pages/Explore/SmartDrawer.tsx
+++ b/src/pages/Explore/SmartDrawer.tsx
@@ -8,6 +8,7 @@ interface SmartDrawerProps {
   title?: string;
   isOpen: boolean;
   onClose: () => void;
+  embedded?: boolean;
   forceNoDrawer?: boolean;
   investigationButton?: React.ReactNode;
 }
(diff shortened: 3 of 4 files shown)


src/exposedComponents/types.ts
```diff
@@ -19,6 +19,7 @@ export interface OpenInExploreTracesButtonProps {
 export interface EmbeddedTraceExplorationState extends SharedExplorationState {
   initialTimeRange: TimeRange;
   onTimeRangeChange?: (timeRange: TimeRange) => void;
+  urlSync?: boolean;
```
joey-grafana, line 22: Where are we toggling this on - another PR?

joey-grafana (changes requested):

src/exposedComponents/types.ts
ifrost: Yep - it's passed in https://github.com/grafana/asserts-app-plugin/pull/1411

joey-grafana: Thanks, wasn't 100% sure if that was the final PR equivalent to this one :)

ifrost:
Not sure if it's something with my setup but clicking in "Add investigation" does nothing in Embedded mode. Works fine in Traces Drilldown, though:


https://github.com/user-attachments/assets/d6672e59-2314-4928-bb5b-01d6c5bf45ee

src/components/App/globalStyles.css
```diff
@@ -0,0 +1,5 @@
+@media screen and (max-width: 1300px) {
```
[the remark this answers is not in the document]
ifrost: The only concern I have if that would affect any existing drawers in a way we don't want to happen 🤔

src/pages/Explore/AttributeFiltersVariable.tsx
```diff
@@ -0,0 +1,69 @@
...
+              origin: this.embedderName,
+            };
+          }
+
+          return filter;
+        });
+
+        // Only update if there are actual changes
+        if (hasChanges) {
+          this.setState({ filters: updatedFilters });
```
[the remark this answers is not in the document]
joey-grafana: Ok we're good to merge for now, but this is something I'd like us to improve in a subsequent PR.

joey-grafana (approved):

PR #472, opened 2025-05-20
Bring back all primary signals

adrapereira:
This PR restores all primary signals by adding a dropdown next to the existing radio button group for "Root spans" and "All spans". When an option from the dropdown is selected it will move to the button group. When it stops being selected the option goes back to the dropdown.

https://github.com/user-attachments/assets/05b7ed0a-bd02-497b-813d-6e9e5544c193

### DIFF
--- a/src/pages/Explore/PrimarySignalVariable.tsx
+++ b/src/pages/Explore/PrimarySignalVariable.tsx
@@ -1,12 +1,14 @@
 import React, { useEffect } from 'react';
 import { CustomVariable, MultiValueVariable, MultiValueVariableState, SceneComponentProps } from '@grafana/scenes';
 import { primarySignalOptions } from './primary-signals';
-import { Icon, RadioButtonGroup, Select } from '@grafana/ui';
+import { Icon, RadioButtonGroup, Select, useStyles2, Text } from '@grafana/ui';
 import { css } from '@emotion/css';
 import { components, DropdownIndicatorProps } from 'react-select';
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from 'utils/analytics';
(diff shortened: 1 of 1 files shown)


alexbikfalvi:
Would it make sense to make the drop down wider so that the descriptions don't wrap only 1-2 words per line? cc @nadinevehling 

Screenshot 2025-05-21 at 9 11 50 AM

ifrost (approved):
Nice 🚀  Agree with Alex - it'd be nice to make the dropdown a bit wider

nadinevehling (changes requested):
Nice, especially because defaults are displayed and secondary functionality is provided on demand.

A few enhancements here:

- Dropdown interaction: 
a) Can we store the primary in local storage? If users rely on a primary signal in the dropdown, can we persevere the action if they selected the last time? 
b) Change the search icon into the opposite dropdown pointing up.

- Bring back the [head to the primary signal dropdown](https://www.figma.com/design/GcixZJYCEqy7CGqEou1Zbt/Private-preview-workflows--Comparison--Filtering?node-id=778-9740&t=tKrgu2EG83TC8Cwo-0). The image shows the head, please disregard the wording "Full traces".
Screenshot 2025-05-22 at 11 35 01

- Please also add analytics to track Rudderstack events for each of the primary signals. This will help us further develop the most used options.

nadinevehling:
Nice, thanks for updating!

Okay, we can try that! It can be irritating over time if users rely on database calls for example. But that's why I pointed out analytics. ;)

Please also add the divider line as well between the "primary signal head" and the other options to clarify users can't select the head and to give some hierarchy and differentiation.

nadinevehling:
> Updated the menu group heading style, is this what you wanted? It matches what we had before migrating to the button group.
> 


Yes, exactly 👍

PR #477, opened 2025-05-29
Embedded mode improvements

joey-grafana:
This PR adds the following improvements in embedded mode;

- Default to trace list tab if there are enough traces as this seems to be the user flow,
- Default to the appropriate metric based on the asserts context.

The equivalent Asserts [PR](https://github.com/grafana/asserts-app-plugin/pull/1492).

### DIFF
--- a/src/exposedComponents/EmbeddedTraceExploration/EmbeddedTraceExploration.tsx
+++ b/src/exposedComponents/EmbeddedTraceExploration/EmbeddedTraceExploration.tsx
@@ -21,7 +21,7 @@ function buildTraceExplorationFromState({
     }
   });
 
-  const exploration = new TraceExploration({ $timeRange, embedded: true, ...state });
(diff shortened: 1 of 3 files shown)


src/components/Explore/TracesByService/Tabs/TabsBarScene.tsx
```diff
@@ -51,6 +52,19 @@ export class TabsBarScene extends SceneObjectBase<TabsBarSceneState> {
...
+        return;
+      }
+
+      // Set the view to traceList if the data is loaded and the traces count is greater than 20
+      if (exploration.state.embedded && dataState.data?.state === LoadingState.Done && tracesCount !== undefined && tracesCount > 20) {
+        hasSetView.current = true;
+        metricScene.setActionView('traceList');
+        return;
+      }
+    }, [dataState.data?.state, tracesCount]);
```
ifrost, line 66: I haven't tested it, but could it inadvertently switch to traceList from some other view when the data is refreshed?

ifrost (approved):

PR #485, opened 2025-06-18
Add namespace to variables

joey-grafana:
Added `tracesDrilldown` namespace to prevent variable collisions. See more on this [here](https://github.com/grafana/scenes/pull/1143) and an alternative approach [here](https://github.com/grafana/scenes/pull/1156).

This PR will act as a way to prevent the variable collisions for now, with a more permanent solution being introduced in either of the above linked PR's.

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -57,6 +57,7 @@
     "localblocks",
     "grafanacloud",
     "OLJCESPC",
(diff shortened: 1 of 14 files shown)


joey-grafana:
Fixed in https://github.com/grafana/scenes/pull/1156

Issue #480, opened 2025-06-12
Filter by label values not escaping values


jfraudeau:
When setting up filters with special characters in values, the values are not escaped resulting in query errors

In the UI:
![Image](https://github.com/user-attachments/assets/3a36c0ef-462a-4b47-b4bf-9ea8a1c7975c)

Payload sent to without escaping, we can see that the quotes are escaped for json but there is no traceql escaping
```json
{
  "queries": [
    {
      "refId": "A",
      "query": "{nestedSetParent<0 && name=\"a\" b && status!=error} | rate() by(status)",
      "queryType": "traceql",
      "tableType": "spans",
      "limit": 100,
      "spss": 10,
      "filters": [],
      "step": "28s",
      "datasource": { "type": "tempo", "uid": "grafanacloud-traces" },
      "serviceMapQuery": "",
      "datasourceId": 10,
      "intervalMs": 30000,
      "maxDataPoints": 64
    }
  ],
  "from": "1749727227291",
  "to": "1749729027291"
}

```

PR #492, opened 2025-06-27
Escape quotes and backslashes (#480)

jfraudeau:
As stated in the linked issue, we don't currently escape filter values when rendering queries. I added a test and implemented escaping as documented in https://grafana.com/docs/tempo/latest/traceql/construct-traceql-queries/#examples-2

I think the logic of only adding quotes around the filter value is not already quoted may break uses cases where a user wants to match a quoted value (if my attribute "name" should be equals to literally "value" quotes included)

I'm not sure if this behavior is needed elsewhere in the app so I only altered it slightly by skipping the quoting only if the value begins AND starts by a quote.

Aside: prettier seems to alter the files I modified quite a bit so I had to format manually, not sure if my setup is broken or not all file adhere to the prettier config defined at the root of the repo.

### DIFF
--- a/.github/workflows/bundle-stats.yml
+++ b/.github/workflows/bundle-stats.yml
@@ -22,4 +22,4 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - uses: grafana/plugin-actions/bundle-size@main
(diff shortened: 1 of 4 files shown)


jfraudeau:
Hi @joey-grafana is grafana open to this kind of outside contribution ? If so is there anything I can do to help move the PR forward ?

PR #505, opened 2025-07-15
Include errors in rate queries

adrapereira:


joey-grafana (commented):
I agree with this approach but I think it will cause issues with the exemplar icons (status=ok will be diamond icon, status=error will be x icon)

joey-grafana:
> It won't because we're not grouping by status anymore so there's only one series, which means only one type of exemplar icon.

Great, I didn't check out the branch to test but that sounds correct. 

Would you mind updating the `groupByStatus` [param](https://github.com/grafana/traces-drilldown/blob/main/src/components/Explore/queries/generateMetricsQuery.ts#L10) to something more appropriate?

Also, there are tests that need updating before we can merge but otherwise looks good!

PR #533, opened 2025-07-25
chore: Send time range in getTagKeys request

joey-grafana:
Sending the time range in getTagKeys request, allows us to use the Tempo datasource setting which sets the range of a tag request to the Tempo language provider.

Companion [PR](https://github.com/grafana/grafana/pull/108691) in Tempo.


joey-grafana:
Bit confused here: the compatibility check is [failing](https://github.com/grafana/traces-drilldown/actions/runs/16522679511/job/46728002688?pr=533) because our module.tsx is using `addLink` and `addComponent` in an incompatible way... for version `12.1.0` of Grafana... which we're not using!! And we specify the exact version  of `12.0.0` in our `package.json`.

Also, I've never seen this check before - must be something new since we updated the packages? :D

PR #554, opened 2025-09-12
feat: Percentiles variable for duration breakdown

adrapereira:
This PR adds a variable to select the percentiles used in the query for the duration breakdown. The default value is still p90 but users can select any number of values from this list: 
- p50
- p75
- p90
- p95
- p99

https://github.com/user-attachments/assets/cc356e1c-f536-421d-9cd3-f2610fb12d96


ifrost (approved):
Works great! There's some issue that after refresh group-by sometimes is not applied but I was it in main so must be something else.

I noticed the UI jumps when selecting a value but I don't see it happening on your video (also happens on selecting group by for me 🤔 ) so it's also not related.

The UI feels a bit confusing: when deselecting last value, P90 is selected and it might not be clear why but it's a good improvement as it is. Adding "default" to value, a tooltip or disabling deselecting last value are some alternatives.

PR #555, opened 2025-09-17
feat: View trace by ID input

adrapereira:
This PR adds a new input field to the header of Traces Drilldown that allows users to input a trace ID and view the trace within the app's drawer. To view the trace simply enter the trace ID and press enter. There's also two handy buttons next to the input:
1. Clear button - clears the input completely
2. Enter button - submits the value and opens the trace

https://github.com/user-attachments/assets/0d3b7825-17e6-4740-b484-25f54dcd7095

Additionally fixed an issue where resizing the drawer when there is another app opened (such as the investigations or assistant app) would incorrectly compute the new widths resulting in the drawer extending over the limits of the window.

### DIFF
--- a/src/components/Explore/TracesByService/Tabs/Spans/SpanListScene.tsx
+++ b/src/components/Explore/TracesByService/Tabs/Spans/SpanListScene.tsx
@@ -105,6 +105,12 @@ export class SpanListScene extends SceneObjectBase<SpanListSceneState> {
               if (nameField?.config?.custom) {
                 nameField.config.custom.cellOptions = options;
               }
+
+              const spanIDField = fields.find((f) => f.name === 'spanID');
+              if (spanIDField?.config?.custom) {
+                spanIDField.config.custom.hideFrom = { viz: true };
+              }
+
               return {
                 ...df,
                 fields,
@@ -190,8 +196,6 @@ export class SpanListScene extends SceneObjectBase<SpanListSceneState> {
                   .setHoverHeader(true)
                   .setOverrides((builder) => {
                     return builder
-                      .matchFieldsWithName('spanID')
-                      .overrideCustomFieldConfig('hidden', true)
                       .matchFieldsWithName('traceService')
                       .overrideCustomFieldConfig('width', 350)
                       .matchFieldsWithName('traceName')


src/components/Explore/panels/TraceViewPanelScene.tsx
```diff
@@ -47,6 +87,23 @@ export class TraceViewPanelScene extends SceneObjectBase<TracePanelState> {
                 component: SkeletonComponent,
               }),
             });
+          } else if (data.data?.state === LoadingState.Error) {
+            this.getTraceErrorMessage(data.data?.error, this.state.traceId)
```
joey-grafana, line 91: Error is deprecated

Screenshot 2025-09-24 at 13 21 12

src/components/Explore/panels/TraceViewPanelScene.tsx
```diff
@@ -23,6 +26,43 @@ export interface TracePanelState extends SceneObjectState {
 }
 
 export class TraceViewPanelScene extends SceneObjectBase<TracePanelState> {
+  private async getTraceErrorMessage(error: any, traceId: string): Promise<string> {
```
joey-grafana, line 29: Should not be using `any` type as it can lead to runtime crashes

src/components/Explore/panels/TraceViewPanelScene.tsx
```diff
@@ -23,6 +26,43 @@ export interface TracePanelState extends SceneObjectState {
...
+        const datasourceUid = getDataSource(getTraceExplorationScene(this));
+        const datasource = await getDataSourceSrv().get(datasourceUid);
+
+        // Check if the datasource has traceQuery.timeShiftEnabled set to true
+        if (datasource && (datasource as any).traceQuery?.timeShiftEnabled) {
+          const timeRange = sceneGraph.getTimeRange(this).state.value;
+
+          // Get timeshift values from datasource configuration
+          const spanStartTimeShift = (datasource as any).spanStartTimeShift || 0;
+          const spanEndTimeShift = (datasource as any).spanEndTimeShift || 0;
```
joey-grafana, line 46: Also should not use `any` (it's returning 0) here as it's actually not getting the shifts at all because they are in the `traceQuery` object!

joey-grafana (changes requested):
Nice work! I see a few things I think we can improve.

- When you click on a trace in the trace list it will open the trace view but not update/remove the text in the trace input. We should remove it so users don't get confused with stale values.
- Let's hide the x and enter icons in the input when there is no text to remove or enter.

src/pages/Explore/Drawer.tsx
```diff
@@ -211,9 +211,15 @@ function useResizebleDrawer(): [
...
-  return `${widthPercent}vw`;
+  const traceExploration = document.getElementById('trace-exploration');
+  if (traceExploration) {
+    const traceExplorationWidth = traceExploration.offsetWidth;
+    const offsetLeft = traceExploration.offsetLeft;
+    const offsetRight = traceExplorationWidth - (clientX - offsetLeft);
+    const widthPercent = Math.min((offsetRight / traceExplorationWidth) * 100, 98).toFixed(2);
+    return `${widthPercent}%`;
+  }
+  return '50%';
```
ifrost, line 222: Is there a case when document.getElementById('trace-exploration'); is not defined 🤔

src/components/Explore/panels/TraceViewPanelScene.tsx
```diff
@@ -23,6 +27,47 @@ export interface TracePanelState extends SceneObjectState {
 }
 
 export class TraceViewPanelScene extends SceneObjectBase<TracePanelState> {
+  private async getTraceErrorMessage(errors: DataQueryError[], traceId: string): Promise<string> {
+    const errorMessage = errors?.[0]?.message || '';
+    const status = errors?.[0]?.status;
+
+    if (status === 404 || errorMessage.toLowerCase().includes('not found')) {
```
ifrost, line 34: Not sure why I'm getting 400, not 404 🤔 

Image

src/components/Explore/panels/TraceViewPanelScene.tsx
```diff
@@ -23,6 +27,47 @@ export interface TracePanelState extends SceneObjectState {
...
+      try {
+        // Get the datasource to check timeShiftEnabled configuration
+        // Ideally this error would be returned by the datasource, but it's not currently supported.
```
ifrost, line 37: It'd be great to have this returned by Tempo indeed! Do we have a task for it?

ifrost (approved):

joey-grafana (approved):

PR #558, opened 2025-09-19
feat: Attributes sidebar

adrapereira:
This PR introduces a new attributes sidebar that will replace the current group by attribute selector. Here's what it can do:

- Supports single or multiple selection mode
  - For breakdown and comparison we'll use single attribute selection
  - For the traces list we'll use the multiple attribute selection mode to add attributes to the table
- Favorite attributes list
  - Currently stored in local storage, we can migrate to platform user storage later
  - Initialized with a set of recommended attributes. These are currently static, defined in code, but we can move to plugin settings later and have an admin define the defaults
  - Users can reorder favorite attributes as they please
  - The first attribute in the favorites list is used as default when going to the Breakdown tab, so no more hardcoded service name default
  - When a user adds an attribute to the filter we automatically select the next attribute in the favorites list to avoid showing a single panel in breakdown.
- Separate span and resource tabs with an additional tab with all attributes
- Search field to easily find attributes
- Allow all mode, useful for the Comparison tab where we can show all attributes in the same page. 

https://github.com/user-attachments/assets/a1a2933d-40c7-4367-847c-8286e490e8ce

### DIFF
--- a/src/components/Explore/AttributesSidebar.tsx
+++ b/src/components/Explore/AttributesSidebar.tsx
@@ -243,7 +243,7 @@ export function AttributesSidebar({
       setDraggedIndex(null);
       setDragOverIndex(null);
     },
-    [draggedIndex, reorderFavorites, filteredAttributes]
+    [draggedIndex, reorderFavorites, filteredAttributes, favoriteAttributes]
   );
 
   const handleScopeChange = (scope: ScopeType) => {
--- a/src/components/Explore/AttributesSidebar.test.tsx
+++ b/src/components/Explore/AttributesSidebar.test.tsx
@@ -3,8 +3,7 @@ import { render, screen, fireEvent } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { AttributesSidebar } from './AttributesSidebar';
 import { SelectableValue } from '@grafana/data';
-import { SceneObject } from '@grafana/scenes';
-import { AdHocFiltersVariable } from '@grafana/scenes';
+import { AdHocFiltersVariable, SceneObject } from '@grafana/scenes';
 
 // Mock the hooks and utilities
 jest.mock('hooks', () => ({
--- a/src/components/Explore/TracesByService/Tabs/Breakdown/AttributesBreakdownScene.tsx
+++ b/src/components/Explore/TracesByService/Tabs/Breakdown/AttributesBreakdownScene.tsx
@@ -112,6 +112,7 @@ export class AttributesBreakdownScene extends SceneObjectBase<AttributesBreakdow
       if (!groupBy || groupBy === 'All' || groupBy === '') {
         model.onChange(favoriteAttributes[0]);
       }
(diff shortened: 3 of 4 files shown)


src/components/Explore/GroupBySelector.tsx
ifrost, line 1: Is GroupBySelector component used anywhere? 🤔

src/components/Explore/AttributesSidebar.tsx
```diff
@@ -0,0 +1,591 @@
...
+    });
+  }, [attributeItems, searchValue, selectedScope, favoriteAttributes]);
+
+  // Select the next favorite attribute if the selected attribute is in the filters (single mode only)
+  useEffect(() => {
+    if (!isMulti && selected && typeof selected === 'string' && currentFilters.includes(selected)) {
+      const currentIndex = filteredAttributes.findIndex((item) => item.value === selected);
+      const nextIndex = currentIndex + 1;
+
+      if (nextIndex < filteredAttributes.length) {
```
ifrost, line 140: I think this actually allows selecting an attributes that is in filters if it's the last one:

Image

ifrost (approved):
This looks fantastic! 





The PR is too big to get into the details so just left a few comments. I don't see any red flags and it's a such a cool feature it'd be nice to get it in and tweak later if needed 👍 

**on UX**

The tabs for sidebar have different style than tabs in the main content. It ight be worth trying to see how it would look if they were using the same style. Also the fact that the container is not having a border doesn't help visually connect tabs with the list (at least to me).

> When a user adds an attribute to the filter we automatically select the next attribute in the favorites list to avoid showing a single panel in breakdown.

I started testing it in the embedded mode where some filters were preselected and I didn't pay attention to what was selected in filters. When I saw a label that is underlined and not selectable but clickable I couldn't figure out why. It'd be nice to have a better visual feedback (a separate section sticked at the top, or a little filter icon with a tooltip, or even just a tooltip on hovering the label).

**on performance**

It's hard to reproduce but when I started switching between single/grid view the UI froze 😱  It was sending thousands of errors like below, and eventually causing a freeze:

Image

I think it's already happening in main, not related to this PR, and could be fixed seprately but please double check if it's not more prominent with this PR.

**on tests**

I'd be good to add some test coverage. There's a lot of features in the side bar and we put them all in one PR so there's always more risk something might be broken.

joey-grafana (commented):
Overall it looks good - great feature 👍  Only one issue and there are a few UX paper cuts.

**Error**

If I select any attribute via add to filters it works fine, but when I select service name add to filters it causes an exception.

Screenshot 2025-10-01 at 08 49 44




**Paper cuts**

- no border around the component like the other panels next to it
- selected tab has a different style to the tabs above (breakdown, comparison etc.)
- height of rows in favourites, all is taller than those in span, resource - makes it a little jarring when moving between tabs

Also no regex search (which people love to ask for) but that's something we can add later!

PR #592, opened 2025-11-11
feat: Mini embeddable Traces Drilldown

joey-grafana:
Exposes a mini embeddable Traces Drilldown for the Assistant. Now when you hover over a link to Traces Drilldown from the Assistant you will see the mini version to get a snapshot look at what you would be navigating to.

You can also click on any of the RED metric panels in mini mode which will navigate you to Traces Drilldown with the appropriate filters added.

Equivalent Assistant [PR](https://github.com/grafana/grafana-assistant-app/pull/2924).

Screenshot 2025-11-11 at 10 35 04


src/components/Explore/queries/generateMetricsQuery.ts
```diff
@@ -1,13 +1,15 @@
+import { getTraceExplorationScene } from 'utils/utils';
 import { ALL, MetricFunction, VAR_FILTERS_EXPR, VAR_DURATION_PERCENTILES_EXPR } from '../../../utils/shared';
+import { SceneObject } from '@grafana/scenes';
 
 interface QueryOptions {
   metric: MetricFunction;
   extraFilters?: string;
   groupByKey?: string;
-  sample?: boolean;
+  sample?: string;
```
adrapereira: Not a fan of having the sample be a free string, there's more potential for mistakes when generating the query. If we must support different sample parameters can you consider an enum? Then we can control what get's added to the query for each enum value in the generateMetricsQuery function.

PR #594, opened 2025-11-13
feat: Find single span errors and high latency on the root cause tab

joe-elliott:
Currently the root cause tab runs a traceql query like this:

`{ <some filters> && status = error } &>> { status = error }`

This finds trees of errors, aggregates them and nicely displays them to the user. Unfortunately, this query returns nothing if there is only a single span with errors in the trace. In that case it displays a confusing blank page. This PR modifies the query to be:

`({ <some filters> && status = error } &>> { status = error }) || ({ <some filters> && status = error })`

This will find both trees of errors and single span errors.

**Before:**

image

**After:**

image

**Other:**

- I added a data source to allow local proxying when doing local dev. I'd be glad to remove if this is not desired.
- I wasn't sure how to test this change. Please advise if I should add a test.
- Applied the changes to all 3 metrics types: Errors, rates, and duration. Same effect for all.

### DIFF
--- a/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
+++ b/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
@@ -366,13 +366,9 @@ function buildQuery(metric: MetricFunction) {
     selectionQuery.length ? `&& ${selectionQuery}` : ''
   }`;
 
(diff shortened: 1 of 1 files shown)


adrapereira (approved):
LGTM!

PR #610, opened 2025-12-08
hackathon-15: create `get-annotations` extension for `adaptive-telemetry-insights`

darrenjaneczek:
- Create a `get-annotations` plugin extension point (function extension) which collects annotations from participating plugins
- Create a `enrichWithAnnotations` behavior and `useAnnotationExtensions` react hook to configure the plugin extensions and execute the request for annotations data frames.
- Apply an alternative annotation mode to "multiLane" annotations on the REDPanel timeseries and histogram panel builders
  - Centralized to a private methodon REDPanel, which could be controlled by a setting
  - Slight change to the histogram panel generator to accept a callback to make further panel config changes

### DIFF
--- a/src/components/Explore/TracesByService/REDPanel.tsx
+++ b/src/components/Explore/TracesByService/REDPanel.tsx
@@ -32,7 +32,7 @@ import {
   getUrlForExploration,
   shouldShowSelection,
 } from '../../../utils/utils';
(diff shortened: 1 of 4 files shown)


joey-grafana:
Annotations have been implemented elsewhere - thank you!

Issue #590, opened 2025-11-04
Time range seeker


joey-grafana:
⭐⭐ Time range seeker
Component at the top of the TD app with a long time range (1d - 7d), heavily sampled to keep performance at a decent level, used to narrow down a time range that may contain an interesting set of data. Main use is for exploration of data.

PR #611, opened 2025-12-09
feat: Time range seeker

adrapereira:
Fixes: #590 

Users of Grafana Traces Drilldown still struggle when trying to discover issues because the plugin triggers multiple queries whenever a time range is selected, which slows down the experience because these requests take a long time for long time ranges. 

This PR introduces a time range seeker under the main metric chart which allows users to easily find spikes, select the time range and drill down on that specific range. This chart is able to show a much larger time range due to a few optimisations:
 - Single request for the time seeker: unlike the main time range, we don't trigger requests for all charts visible in the page, we perform a single request for a metrics query
 - Sampling: the query generated requests that Tempo samples the result, which may result in lower fidelity of results but the response returns significantly faster. Lower fidelity is fine for this case because in longer time ranges we're looking for spikes instead of granular values.
 - 24h batches: By requesting the data in batches we can show a large time range without worrying about going over the range limit defined in Tempo. Users can configure the batch range in the plugin's configuration page, which should be slightly smaller than the max range allowed in Tempo. 

Requires toggle to enable https://github.com/grafana/traces-drilldown/pull/611/changes/27d9df11d3db5a1e1daaf2bb451f282754f43957

Demo:

https://github.com/user-attachments/assets/939dbb0d-fa8c-461a-b346-2b0e9f4692ab

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,10 +67,10 @@
     "prettier": "3.6.2",
     "replace-in-file-webpack-plugin": "1.0.6",
     "sass": "1.89.2",
-    "sass-loader": "16.0.5",
-    "semver": "7.7.2",
+    "sass-loader": "16.0.7",
+    "semver": "7.7.4",
     "style-loader": "4.0.0",
-    "swc-loader": "0.2.6",
+    "swc-loader": "0.2.7",
     "terser-webpack-plugin": "5.3.14",
     "ts-node": "10.9.2",
     "typescript": "5.8.3",
@@ -107,7 +107,7 @@
     "rxjs": "7.8.2"
   },
   "resolutions": {
-    "nwsapi": "2.2.22",
+    "nwsapi": "2.2.23",
     "@tootallnate/once": ">=3.0.1",
     "schema-utils/ajv": "8.18.0",
     "rxjs": "^7.8.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6903,10 +6903,10 @@ nth-check@^2.0.1:
   dependencies:
     boolbase "^1.0.0"
 
-nwsapi@2.2.22, nwsapi@^2.2.2:
-  version "2.2.22"
-  resolved "https://registry.yarnpkg.com/nwsapi/-/nwsapi-2.2.22.tgz#109f9530cda6c156d6a713cdf5939e9f0de98b9d"
-  integrity sha512-ujSMe1OWVn55euT1ihwCI1ZcAaAU3nxUiDwfDQldc51ZXaB9m2AyOn6/jh1BLe2t/G8xd6uKG1UBF2aZJeg2SQ==
+nwsapi@2.2.23, nwsapi@^2.2.2:
+  version "2.2.23"
+  resolved "https://registry.yarnpkg.com/nwsapi/-/nwsapi-2.2.23.tgz#59712c3a88e6de2bb0b6ccc1070397267019cf6c"
+  integrity sha512-7wfH4sLbt4M0gCDzGE6vzQBo0bfTKjU7Sfpqy/7gs1qBfYz2vEJH6vXcBKpO3+6Yu1telwd0t9HpyOoLEQQbIQ==
 
 object-assign@^4.1.0, object-assign@^4.1.1:
   version "4.1.1"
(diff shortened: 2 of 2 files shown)


joey-grafana (changes requested):
Very cool feature 👍 

Have you seen the new Timeseries panel functionality where you can click and drag the x-axis of any Timeseries (which already uses UPlot) panel to pan time backwards or forwards precisely - seems like there is a lot of overlap with this PR, we could probably save ourselves complex code :)

What happens if the user only has trace metrics for the last 24 hours or even less in some cases if they have huge amounts of data.. how would the seeker behave?

I scrolled left on my magic mouse and somehow scrolled out a few years! And the browser got stuck and I had to close the tab!

Screenshot 2025-12-16 at 13 09 07

When I move the seeker range there are a whole lot of queries run! I don't think they're all necessary. Can we tidy this up and reduce queries?

Screenshot 2025-12-16 at 13 11 46

src/pages/Explore/TraceExploration.tsx
```diff
@@ -685,5 +697,13 @@ function getStyles(theme: GrafanaTheme2, embedded?: boolean, embeddedMini?: bool
       display: 'flex',
       gap: theme.spacing(1),
     }),
+    filtersColumn: css({
```
joey-grafana: unused

src/components/Explore/seeker/TimeSeeker.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <div className={cx(styles.wrapper)}>
+        <TimeSeekerControls />
+        <TimeSeekerChart />
+      </div>
+    </TimeSeekerProvider>
+  );
+};
+
+// Re-export types for convenience
+export type { TimeSeekerProps } from './types';
```
joey-grafana: nit: unnecessary

src/components/Explore/seeker/TimeSeekerChart.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+      background: { primary: '#fff' },
+      border: { weak: '#ccc' },
+      primary: { shade: '#007bff' },
+    },
+    visualization: { getColorByName: (name: string) => name },
+  }),
+  useStyles2: (fn: Function) => fn({
+    colors: { background: { primary: '#fff' }, primary: { shade: '#007bff' } },
+    spacing: () => '8px',
+  }),
```
joey-grafana: Can prob remove this

src/components/Explore/seeker/TimeSeekerControls.tsx
```diff
@@ -0,0 +1,154 @@
...
+    requestAnimationFrame(() => {
+      setTimeout(() => {
+        const button = container.querySelector('button') as HTMLButtonElement;
+        if (button) {
+          button.focus();
+          button.click();
+          return;
+        }
+      }, 100);
+    });
```
joey-grafana: Timeouts/requestAnimationFrame are a bit messy and not sure why it's needed, could prob just write

```
 if (button) {
    button.click();
 }
```

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
```diff
@@ -0,0 +1,211 @@
...
+
+        setVisibleRange({ from: newFrom, to: newTo }, true);
+      };
+
+      // Attach wheel zoom to uPlot overlay
+      const over = u.root.querySelector('.u-over') as HTMLElement;
+      if (over && wheelListenerRef.current) {
+        over.addEventListener('wheel', wheelListenerRef.current, { passive: false });
+
+        (u as any)._cleanupWheelZoom = () => {
```
joey-grafana: Could we remove uses of any from this file please?

src/components/Explore/seeker/TimeSeekerContext.tsx
```diff
@@ -0,0 +1,595 @@
...
+  // Refs
+  // -------------------------------------------------------------------------
+  const uplotRef = useRef<uPlot | null>(null);
+  const wheelListenerRef = useRef<((e: WheelEvent) => void) | null>(null);
+  const applyRelativeContextWindow = useRef<string | null>(null);
+  const lastDashboardRange = useRef<AbsoluteTimeRange>({ from: dashboardFrom, to: dashboardTo });
+
+  // Interaction tracking - consolidates multiple boolean flags
+  const interactionMode = useRef<InteractionMode>('idle');
+  const suppressNextDashboardUpdate = useRef(false);
```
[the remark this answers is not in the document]
adrapereira: These flags work together to avoid infinite loops in rendering or updating the time range by avoiding multiple actions being performed at the same time (`interactionMode`) or skipping certain updates depending on the action origin. This is already simplified when compared to the initial PR version. The `suppressNextDashboardUpdate` variable existed in the initial component POC which would have been used in dashboards but since this is an app plugin I renamed it to `suppressNextTimeRangeUpdate` to make it clearer.

knylander-grafana:
> @knylander-grafana FYI we'll need docs for this feature once released. Let me know if you need anything from us!

What's our time frame? I want to make sure I have the docs on the schedule for this.

Issue #589, opened 2025-11-04
Exceptions tab version 2


joey-grafana:
⭐⭐⭐⭐ Exceptions tab version 2
Users want a more streamlined experience when it comes to viewing their exceptions in TD. A new and improved version has been designed which will be the base for the improvements here.

PR #619, opened 2026-01-09
feat: Exceptions Tab v2

joey-grafana:
Upgrades the Exceptions tab to build a better design and visual hierarchy. 

Adds include/exclude buttons to make updating the query easier.

Part of https://github.com/grafana/traces-drilldown/issues/589

Screenshot 2026-01-09 at 12 45 54

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -75,6 +75,7 @@
     "automerge",
     "nwsapi",
     "adaptivetraces",
-    "tailsampling"
+    "tailsampling",
+    "sparklines"
   ]
 }


src/components/Explore/TracesByService/Tabs/Exceptions/ExceptionsTable.tsx
```diff
@@ -0,0 +1,349 @@
...
+  };
+
+  return (
+    <div className={styles.container}>
+      <table className={styles.table}>
+        <thead>
+          <tr className={styles.headerRow}>
+            <th className={styles.headerCell}>
+              <span>Exception Details</span>
+              <Tooltip content="Exception type, message, service, and last seen timestamp. Use the +/- buttons to include or exclude exception types, or click on type or message to filter.">
```
adrapereira: We don't have +/- buttons

adrapereira (approved):
Love it! Approving to unblock you but left a comment and have a question.

Should we have smarter grouping for the exceptions? In this example the exceptions have the same type, but since they don't have the same message (because of a timestamp) we see a lot of lines where it would have better if they were grouped. 

Image

cedricziel:
> Should we have smarter grouping for the exceptions

We should. - Same as @adrapereira , I would love to experiment with smarter grouping.

ifrost (approved):
Looks sooo much better!

PR #624, opened 2026-01-13
feat: Support include / exclude for panels

joey-grafana:
Adds support for include & exclude buttons for filters in breakdown and comparison panels.


adrapereira (commented):
Can we use the same styling from Logs Drilldown?

Image

Issue #567, opened 2025-10-20
Exceptions: Streamline analysis workflow


cedricziel:
## Status Quo

When finding a trending exception, the user can select it and it takes them to a screen that has all the traces that contain it.

I then need to select a sample trace and find the exception again in the span events.

This is unintuitive and time-consuming.

## Desired state

When finding an interesting exception, i should be able to zoom into it and
* see a detail-screen that highlights the most important information (stacktrace, type, comparison by attributes) so i can dive into analysis straight away AND look at samples of that exception

joey-grafana:
Completed in https://github.com/grafana/traces-drilldown/pull/627

PR #627, opened 2026-01-20
feat: Enhanced exceptions tab

joey-grafana:
Creates an enhanced version of the Exceptions Tab.

This version includes a dropdown accordion with a full stack trace, comparison view and traces results table in once place, satisfying the design proposed in Figma and the requirements [here](https://github.com/grafana/traces-drilldown/issues/567).

Further, the comparison shows the distribution of values for a given attribute and each value in the trace results table can be clicked to open up the trace view drawer.

Completes https://github.com/grafana/traces-drilldown/pull/589


joey-grafana:
Thanks @nicwestvold, I've pushed updates so should be good to approve!

PR #648, opened 2026-02-10
chore: Remove spellcheck

joey-grafana:
This was just getting in the way, failing PR's & never really taken seriously.


bohandley (approved):
I like the idea of a spellcheck but happy to remove something that results in flaky tests and paper cuts on merging PRs. 

Grafana has so many custom and inconsistent words, data sources vs datasources for example, I am happy to see this removed and maybe look in a more mature and less flaky spell check in the future.

Or just add this to the bottom of our PR templates 😁:

- [ ] PR has been checked for spelling by one human and one LLM

joey-grafana:
Thanks fellas yeah honestly it's interesting because it sounds like a helpful tool but for us it just has been a thing where we have to update the cspell file when we randomly add a new word somewhere but it has not really helped with spelling.

PR #685, opened 2026-03-26
feat: add Knowledge Graph insights annotations to timeseries

rdubrock:
## Summary
- Adds Knowledge Graph annotation overlays to timeseries panels in Explore Traces, gated behind the `kgAnnotationsInExploreTraces` feature flag
- Creates `KgAnnotationToggle` scene control (Insights on/off switch) and `kgAnnotations` utility that builds `AnnotationsDataLayer` instances per severity level (critical/warning/info)
- Integrates into `TraceExploration` constructor, setting `$data` layer set and adding the toggle to controls

## How it works
When the feature flag is enabled and the KG datasource (`grafanacloud-knowledgegraph`) is available, the plugin creates three annotation data layers (one per severity). These overlay colored time regions on timeseries panels. An "Insights" toggle in the header controls lets users enable/disable the annotations.

Entity type is `Service`, and the service name is extracted from `initialFilters` (`resource.service.name`).

Part of https://github.com/grafana/asserts-app-plugin/issues/2922 (epic: https://github.com/grafana/asserts-app-plugin/issues/2919)

Depends on https://github.com/grafana/asserts-app-plugin/pull/3062

## Test plan
- [ ] Enable `kgAnnotationsInExploreTraces` feature flag
- [ ] Verify KG datasource is configured (`grafanacloud-knowledgegraph`)
- [ ] Navigate to Explore Traces with a service filter applied
- [ ] Confirm "Insights" toggle appears in header controls
- [ ] Confirm colored annotation regions appear on timeseries panels
- [ ] Toggle Insights off and verify annotations disappear
- [ ] With feature flag disabled, verify no toggle or annotations appear

### Steps to test
1. Ensure kgAnnotationsInExploreTraces is enabled (set in GF_FEATURE_TOGGLES_ENABLE in docker-compose)
2. Open Explore Traces
3. Enable the toggle -- it appears in the header controls once filters are set
4. Path A -- service name known: Set resource.service.name = <your-service>. This uses a direct KG entity lookup.
5. Path B -- fromLabels fallback: Set one or more other equality filters (e.g. resource.k8s.namespace.name = production) without resource.service.name. The KG resolves the entity from those labels + the trace datasource UID.
6. Expected: Colored time regions overlay timeseries panels -- red = critical, yellow = warning, blue = info
7. Clear all filters -- layers should be removed and toggle should disappear

| Traces | Insights |
| :--- | :---: |
| Screenshot 2026-04-08 at 3 45 06 PM | Screenshot 2026-04-08 at 3 45 15 PM |

🤖 Generated with [Claude Code](https://claude.com/claude-code)
Editted by @bohandley


bohandley (commented):
Looking good!  

I see you implemented both the service name KG lookup and the fromLabels lookup. Would we ever want to send more labels when we have the service name?

Other things to consider:
- This PR feature toggle also needs migrated open feature https://github.com/grafana/metrics-drilldown/pull/1173/changes#r3041997803
- I see the warning color is yellow here as well, metrics drilldown is orange
- Could use some tests :)

Other than that, LGTM!

rdubrock:
> I see the warning color is yellow here as well, metrics drilldown is orange

Ah that's a mistake by me. They are yellow everywhere else, I will follow up with a PR to metrics drilldown

Screenshot 2026-04-07 at 08 13 26

Will get cracking on the other items!

rdubrock:
> Would we ever want to send more labels when we have the service name?

They _can_ help depending on the label. Because entities with the name "foo" running in different environments are considered distinct entities. The "foo" in prod is _not_ the same as "foo" in dev. So it's possible to narrow further than just by name. That's the only scenario. But I also don't want to implementation detail about how to search the graph embedded in consuming plugins, my dream is that the fromLabels approach is just fire and forget. The datasource should be able to figure out what to do with them.

bohandley (commented):
LGTM! This is great work and thank you for the changes @rdubrock!

Annotations insights show up by filtering by resource.service.name or by choosing a combination of labels that can match an entity.

I'm holding off on approving to give time for the traces folks to review, but it should be soon. Thank you!

Image
Image

adrapereira (approved):
This is great @rdubrock, great job!

Couple of things:
- I can still see the old insights button and the insights row under the graph, not sure if it's just my setup but it would be great to not show them when the flag is enabled. 
- The header is getting a little too busy, not expecting any action in this PR but something we must take into account for any further changes.

PR #713, opened 2026-04-14
chore: Use OpenFeature for all feature flags

joey-grafana:
Migrates all remaining feature toggles to fully support OpenFeature and removes the fallback to config.featureToggles.

See https://github.com/grafana/grafana/blob/main/contribute/feature-toggles.md for more details.


nicwestvold (approved):
lgtm!

MattIPv4 (commented):
👀 Up to you, but instead of mocking the React hooks and the OFREPWebProvider, you may want to use OpenFeature's built-in test support in React, which'll let you exercise a bit more of the OpenFeature logic in tests: https://openfeature.dev/docs/reference/sdks/client/web/react#testing

PR #728, opened 2026-04-24
fix: ensure AttributesSidebar displays correctly on smaller screens

### LINKED ISSUE #613: Attributes sidebar doesn't scale well for smaller screens
For smaller screens the attributes sidebar is too small so users can't see the list of attributes and select/deselect them. 

Image

adrapereira:
### Problem

On viewports shorter than ~850px, `height: calc(100vh - 550px)` collapses the tab content area too small for the `AttributesSidebar` to render usefully. The sidebar header could be compressed out of view, and the `ByFrameRepeater` panel body had no overflow scrolling.

### Changes

- **`AttributesSidebar`** — added `minHeight`, `alignSelf: 'stretch'`, and `overflow: 'hidden'` to the container; added `flexShrink: 0` to the header to prevent it from being compressed.
- **Breakdown / Comparison / Spans scenes** — added `minHeight: '500px'` to the `content` area so there is always a usable floor on small screens.
- **`ByFrameRepeater`** — added `overflow: 'auto'` to the panel body so it scrolls instead of clipping content.

### Test plan
- [ ] Open Breakdown / Comparison / Spans tabs at a reduced viewport height (< 850px) and verify the sidebar is fully visible and scrollable.
- [ ] Verify the sidebar header is not compressed at any viewport height.
- [ ] Verify no regressions on normal screens.

Fixes #613

### DIFF
--- a/src/components/Explore/AttributesSidebar.tsx
+++ b/src/components/Explore/AttributesSidebar.tsx
@@ -4,7 +4,7 @@ import React, { useMemo, useState, useEffect, useCallback } from 'react';
 import { GrafanaTheme2, SelectableValue } from '@grafana/data';
 import { TabsBar, Tab, Input, Icon, IconButton, useStyles2, Badge, Checkbox, Button, useTheme2 } from '@grafana/ui';
 import { t, Trans } from '@grafana/i18n';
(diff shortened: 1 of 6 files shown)


joey-grafana (approved):
LGTM but please use a const for `500px` as it's repeated 4 times and if we change this in the future it will be easier to change once

PR #731, opened 2026-04-24
fix: Scroll to top when opening a linked span

### LINKED ISSUE #595: Trace is not visible after scrolling root causes
This is likely replicable elsewhere but I found it on the Root Cause tab. To reproduce:

1) Choose "All spans"
1) Choose "Span rate"
1) Click "service structure"

These exact steps aren't necessary. They just tend to create a lot of content in the structural tab which is what creates the issue.

Image

4) Scroll to the bottom 
4) Click "View linked span"

Image

6) Note that the trace loads but the user is staring at a blank page! The trace exists but is way at the top and we are scrolled to the bottom:

Image

7) Scroll to the top and see the trace:

Image

No suggestions on a fix :) but it took me by surprise at first. I thought Tempo was struggling to return the trace.

adrapereira:
## Scroll to top when opening trace drawer from structure scene

When a user clicks a linked span in the Structure tab, the trace drawer opens but the page may be scrolled down, leaving the drawer header out of view. This change ensures the window scrolls back to the top whenever the drawer is opened.

**Changes:**
- `Drawer.tsx` — adds a `useEffect` that calls `window.scrollTo({ top: 0, behavior: 'instant' })` on mount, so the drawer header is always visible when the drawer opens.
- `StructureScene.tsx` — adds `maxHeight: 100vh` and `overflow: auto` to the structure panel container, constraining its height so the page doesn't accumulate excessive scroll depth in the first place. Also includes minor JSX formatting cleanup.

Fixes #595

### DIFF
--- a/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
+++ b/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.tsx
@@ -255,8 +255,16 @@ export class StructureTabScene extends SceneObjectBase<ServicesTabSceneState> {
       case 'rate':
         description = (
           <>
-            <div><Trans i18nKey="structure-scene.rate-description">Analyse the service structure of the traces that match the current filters.</Trans></div>
-            <div><Trans i18nKey="structure-scene.rate-panel-info">Each panel represents an aggregate view compiled using spans from multiple traces.</Trans></div>
+            <div>
+              <Trans i18nKey="structure-scene.rate-description">
+                Analyse the service structure of the traces that match the current filters.
+              </Trans>
+            </div>
+            <div>
+              <Trans i18nKey="structure-scene.rate-panel-info">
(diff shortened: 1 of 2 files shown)


nicwestvold (approved):
Notes from the AI review that may or may not be helpful.

- The `window.scrollTo({ top: 0, behavior: 'instant' })` runs on every drawer mount, not just when opened from the Structure tab. This is likely fine as a general UX improvement, but worth confirming that scrolling to top on all drawer opens (e.g., from other tabs or contexts) is the desired behavior.
- Adding `maxHeight: 100vh` and `overflow: auto` to the structure panel introduces a nested scrollbar (panel scrolls independently from the page). Verify this doesn't create awkward scroll interactions, especially when the panel content is just slightly taller than the viewport.

joey-grafana:
Closing as there are still a number of issues. Can reopen later when we get a chance.

PR #734, opened 2026-04-27
fix: gate deploy on CI success via `workflow_run`

nicwestvold:
### ✨ Description

Gates the CD workflow on CI success so builds are no longer deployed to dev/ops before tests pass.

### 📖 Summary of the changes

Previously, CI and CD both triggered independently on push to `main` and ran in parallel — a failing CI run would not prevent CD from deploying.

- Replaced `push` trigger with `workflow_run` (fires only after the `CI` workflow completes on `main`)
- Added `if` guard: only proceeds when `workflow_run.conclusion == 'success'`
- Added `workflow_dispatch` trigger for manual re-deploys without re-running CI
- Updated `branch` and `plugin-version-suffix` refs to use the `workflow_run` event payload, with ternaries to handle both trigger paths

### 🧪 How to test?

1. Merge to `main` and verify the CD workflow only starts after CI completes successfully
2. Trigger a manual deploy via the "Run workflow" button in GitHub Actions and verify it runs without needing CI

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -3,7 +3,7 @@ run-name: Deploy main to dev,ops by @${{ github.actor }}
 
 on:
   workflow_dispatch:
-  workflow_run:
+  workflow_run: # zizmor: ignore[dangerous-triggers] only triggers from CI on main, no untrusted input
     workflows: ["CI"]
     types: [completed]
     branches: [main]
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,7 +16,7 @@ permissions:
 
 jobs:
   CD:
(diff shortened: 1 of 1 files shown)


shelldandy (approved):
Looks solid to me, let's test it

.github/workflows/cd.yml
```diff
@@ -14,14 +16,15 @@ permissions:
 
 jobs:
   CD:
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
```
joey-grafana:
```suggestion
    if: ${{ github.event_name == 'workflow_dispatch' || (github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push' && github.event.workflow_run.head_repository.full_name == github.repository }}
```

Also checks

- triggering CI event was push

- triggering run came from the same repository

joey-grafana (approved):

PR #744, opened 2026-05-01
feat: Enhance Embedded Traces Drilldown

joey-grafana:
### ✨ Description

Adds the ability to hide rate/error/duration panels in the embedded mode.

This PR also fixes an issue with the comparison tab where `baseline` or `selection_total` values may be wrapped in quotes.

### 🧪 How to test?

Checkout equivalent [Faro branch](https://github.com/grafana/app-o11y-kwl/pull/3273) and use plugin graft to project your local changes onto Frontend O11y in ops.

Screenshot 2026-05-01 at 07 28 03

Screenshot 2026-05-14 at 09 41 19


codecapitano:
Tested with graft on this branch works as epxected but i can still see the embedded time picker rendering in all thre views where we use the taces drilldown.

The "hide" is gated by `initialOrFilters`, so hosts that pass only `initialFilters` (or none) always get two pickers.

Could we widen the condition (e.g. hide whenever `embedded`) or add an explicit `hideTimePicker` prop?

image

codecapitano (approved):

PR #754, opened 2026-05-12
fix: syncing of y axis

### LINKED ISSUE #753: Breakdown panels in Grid view all use the same small y-axis scale, clipping bars in panels with larger values
On the Traces Drilldown Breakdown tab, when I select an attribute that produces multiple panels in Grid view (for example, resource.service.name with many services), every panel is rendered with the same y-axis range. That shared range matches the smallest panel's values, so any panel whose actual values are larger gets clipped — the bars saturate at the top of the chart and look like flat blocks:

Screenshot (bad):
Image

Screenshot (bad):
Image

If I switch the same breakdown to Single view (one chart at a time), the y-axis is sized correctly for that chart and nothing is clipped. The problem only appears when multiple panels are shown side by side:

Screenshot (good):
Image

Grafana Traces Drilldown v2.0.3 (Last update: 2026-04-30 (8d7e2cf))
Grafana: docker.io/grafana/grafana:12.4.2
Tempo: docker.io/grafana/tempo:2.10.1
Browser: 
- Microsoft Edge Version 148.0.3967.54 (Official build) (64-bit)
- Chrome Version 147.0.7727.138 (Official Build) (64-bit)

joey-grafana:
### ✨ Description

Fixes: https://github.com/grafana/traces-drilldown/issues/753

Fix syncing of y axis.

### 🧪 How to test?

Check breakdown tab that each y axis is relative to each other but with different values.

### DIFF
--- a/src/components/Explore/behaviors/syncYaxis.tsx
+++ b/src/components/Explore/behaviors/syncYaxis.tsx
@@ -7,16 +7,18 @@ export function syncYAxis() {
     const eventSub = vizPanel.subscribeToEvent(EventTimeseriesDataReceived, (event) => {
       const series = event.payload.series;
 
-      // Start at -Infinity so the first finite sample becomes the running max; if nothing
-      // numeric is found, globalMax stays non-finite and we skip updateTimeseriesAxis below.
-      let globalMax = Number.NEGATIVE_INFINITY;
+      let globalMax: number | null = null;
       series?.forEach((s) => {
+        // Skip field 0 (typically time); only numeric value fields contribute to the Y max.
         s.fields.slice(1).forEach((f) => {
-          globalMax = Math.max(globalMax, maxFiniteInFieldValues(f.values));
+          const fieldMax = maxFiniteInFieldValues(f.values);
+          if (fieldMax !== null) {
+            globalMax = globalMax === null ? fieldMax : Math.max(globalMax, fieldMax);
+          }
         });
       });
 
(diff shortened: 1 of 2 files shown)


src/components/Explore/behaviors/syncYaxis.tsx
```diff
@@ -24,6 +29,17 @@ export function syncYAxis() {
...
 
+function maxFiniteInFieldValues(values: ArrayLike<unknown>): number {
+  let m = Number.NEGATIVE_INFINITY;
+  for (let i = 0; i < values.length; i++) {
+    const v = values[i];
+    if (typeof v === 'number' && Number.isFinite(v)) {
+      m = Math.max(m, v);
+    }
+  }
+  return m;
```
matyax: I've recently seen Cursor generate single letter variables and I don't think this has any positive side; rather, it hurts readability.

```suggestion
  let max = Number.NEGATIVE_INFINITY;
  for (let i = 0; i < values.length; i++) {
    const value = values[i];
    if (typeof value === 'number' && Number.isFinite(value)) {
      min = Math.max(max, value);
    }
  }
  return max;
```

Another shorter alternative without a loop:


```suggestion
  return  Math.max(...values.filter(value => Number.isFinite(value)))
```

Lastly, if the list is empty or it doesn't contain numbers, the max value will be negative infinity.

matyax (dismissed):

matyax (approved):

PR #756, opened 2026-05-15
feat: Exceptions tab grouping

joey-grafana:
### ✨ Description

Groups similar exception messages in the Exceptions tab so noisy variants (URLs, IDs, numbers, long messages with shared prefix/suffix) collapse into one row. Include/exclude and accordion drill-down use regex filters when the displayed message contains normalization placeholders.

How exceptions are grouped:

- Normalize each exception.message: collapse whitespace, then replace volatile bits with fixed tokens like < url >, < uuid >, < ip >, < hex >, < timestamp >, < id >, < num >.
- Group key = exception.type (if present) + first 20 characters of the normalized text + last 20 + full normalized length. Same key → one table row.
- Aggregate occurrence counts and time series into that row; keep every distinct raw message in groupedMessages for the tooltip.

Rows are then sorted by occurrence count (desc).

### 🧪 How to test?

1. Select errors RED metric
2. Selection exceptions tab
3. Observe exceptions are now grouped

I don't want to share an image because most places seem to have exception info I'd rather not share!! Msg me for a url example if you like. I'm testing with an App O11y env.

### DIFF
--- a/src/components/Explore/actions/AddToFiltersAction.tsx
+++ b/src/components/Explore/actions/AddToFiltersAction.tsx
@@ -96,7 +96,13 @@ function getStyles(theme: GrafanaTheme2) {
   };
 }
 
-export const addToFilters = (variable: AdHocFiltersVariable, label: string, value: string, operator: '=' | '!=' = '=', append = false) => {
+export const addToFilters = (
+  variable: AdHocFiltersVariable,
+  label: string,
+  value: string,
+  operator: '=' | '!=' | '=~' | '!~' = '=',
+  append = false
+) => {
   // TODO: Replace it with new API introduced in https://github.com/grafana/scenes/issues/1103
   // At the moment AdHocFiltersVariable doesn't support pushing new history entry on change
   history.pushState(null, '');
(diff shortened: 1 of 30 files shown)


src/components/Explore/actions/AddToFiltersAction.tsx
```diff
@@ -96,7 +96,13 @@ function getStyles(theme: GrafanaTheme2) {
   };
 }
 
-export const addToFilters = (variable: AdHocFiltersVariable, label: string, value: string, operator: '=' | '!=' = '=', append = false) => {
+export const addToFilters = (
+  variable: AdHocFiltersVariable,
+  label: string,
+  value: string,
+  operator: '=' | '!=' | '=~' | '!~' = '=',
```
matyax, line 103: This construction `'=' | '!=' | '=~' | '!~' = '='`, where it defaults to '=' felt like hieroglyphs.

matyax (approved):
Couldn't find a way to see this with data, but it works and the code LGTM.

PR #789, opened 2026-06-17
chore: Remove homepage

joey-grafana:
### ✨ Description

chore: Remove homepage and legacy code that is no longer used.


shelldandy (approved):
great cleanup thanks!

PR #827, opened 2026-07-30
fix: Scroll to trace in structure tab

### LINKED ISSUE #595: Trace is not visible after scrolling root causes

joey-grafana:
### ✨ Description

Fixes: https://github.com/grafana/traces-drilldown/issues/595

Fixes scroll to trace in structure tab, where previously you could click `View Linked Span` for a trace in the structure tab, which would open the trace drawer but you would not see the trace as you needed to scroll to top.

### DIFF
--- a/src/pages/Explore/Drawer.tsx
+++ b/src/pages/Explore/Drawer.tsx
@@ -84,7 +84,8 @@ export function Drawer({
       open={true}
       onClose={onClose}
       placement="right"
-      getContainer={'#trace-exploration'}
+      // Portal to body with position:fixed so scroll inside #trace-exploration cannot hide the drawer.
+      getContainer={() => document.body}
       className={styles.drawerContent}
       rootClassName={styles.drawer}
       classNames={{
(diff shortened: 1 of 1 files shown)


L2D2Grafana (approved):
LGTM

It does go over the grafana header and docker nav now, but that's probably fine.

Image

in Ops

Image

PR #830, opened 2026-07-31
chore: Remove unused dependencies

joey-grafana:
### ✨ Description

chore: Remove unused dependencies

Also drop stale allowBuilds / override entries for packages no longer in the dependency tree (esbuild, rolldown, protobufjs, @tootallnate/once).

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -16,8 +16,6 @@ blockExoticSubdeps: true
 strictDepBuilds: true
 
 allowBuilds:
-  esbuild: true
-  rolldown: true
   unrs-resolver: true
 
 overrides:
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -16,8 +16,6 @@ blockExoticSubdeps: true
 strictDepBuilds: true
 
 allowBuilds:
-  esbuild: true
-  rolldown: true
   unrs-resolver: true
```
matyax, line 19: Do we want these changes? These are not in the description AFAIK.

joey-grafana: Yes — intentional cleanup. esbuild / rolldown aren’t in the lockfile anymore, so the allowBuilds entries (and related overrides) were unused. I’ll update the description to call that out :)

matyax (approved):

PR #834, opened 2026-08-04
fix(seeker): keep relative range on refresh

### LINKED ISSUE #815: [BUG]: refreshing makes the time range absolute
**Describe the bug**
When I press refresh, the time range becomes absolute.

**To Reproduce**

1. Go to Traces Drilldown
2. The time picker should default to something relative, e.g. Last 30 minutes
3. Click on refresh
4. The time range picker now shows an absolute time range

**Expected behavior**
Time range should be unmodified when I press refresh. Drilldown should fetch fresh data so I can tail incoming traces.

**Environment**
Traces Drilldown 2.1.0+2788a81

**Screenshots**

https://github.com/user-attachments/assets/2ed76bac-4ee2-4938-9816-371824679e93

**AI disclosure** (optional)
See the [Generative AI Contribution Policy](https://github.com/grafana/traces-drilldown/blob/main/docs/genai.md).

- [ ] This issue was substantially generated with AI assistance.

hugohaggmark:
**What is this feature?**

Fixes so a relative time range stays relative after a refresh instead of flipping to an absolute one (the exact same bug happens if the user chooses an auto refresh value in the TimePicker). The time seeker redraws its selection by calling `setSelect` programmatically, and that was treated as a user drag, so it pushed the seeker's absolute window into the main time picker. It now only reacts to real user selections (`isUserSelection`).

So a relative range like `Last 30 minutes` stays relative after refreshing.

Fixes #815

This only affects users with the `tracesDrilldownTimeSeeker` feature toggle enabled, since the time seeker is behind it.

I'm new to this codebase, so I've tried to cover all the cases manually, but someone with more experience should test it out. 🤞

A small test for this could be added to `e2e/time-seeker.spec.ts`, but that suite is commented out until Grafana 13+ is released.


src/components/Explore/seeker/useTimeSeekerChartConfig.ts
hugohaggmark, line 96: maybe all these guards can be removed in the future if the `isUserSelection` function works correclty?

src/utils/utils.test.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('isUserSelection', () => {
+  it('should return true when a real user performs a x-drag with a mouse event and a selection', () => {
+    const u = stubUplot({ event: new MouseEvent('mousedown'), dragX: true, left: 10, width: 50 });
+    expect(isUserSelection(u)).toBe(true);
+  });
+
+  it('should return true even when the selection is anchored at 0 (not null)', () => {
+    const u = stubUplot({ event: new MouseEvent('mousedown'), dragX: true, left: 0, width: 0 });
+    expect(isUserSelection(u)).toBe(true);
+  });
```
hugohaggmark: the existing code (`if (xDrag && u.select.left != null && u.select.width != null)`) accepted `left` and `width` values of `zero` so I didn't change that in this PR, but we might want to think about if this is correct?

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
joey-grafana: isUserSelection only sees uPlot state (drag.x + cursor.event). It doesn’t know about our overlay pan/drag, or suppressions from setVisibleRange(..., true) — those still need interactionMode / isProgrammaticSelect / skipNextSelectUpdate. Fine to revisit later if we simplify those paths.

src/utils/utils.test.ts
joey-grafana: uPlot itself only commits a select when width > 0 || height > 0, so a follow-up requiring width > 0 in isUserSelection would probably be more correct. 

You can do that here or in another PR. I think a lot of this feature was vibe coded, so would ideally like another pass (I already had a go) at simplifying it and removing bloat.

joey-grafana (dismissed):

hugohaggmark:
I added the width check and changed the isUserSelection call to an early exit instead, let me know what you think 👍

src/utils/utils.test.ts
hugohaggmark: Fixed in https://github.com/grafana/traces-drilldown/pull/834/commits/f6fd466d7a4c31b41314d52a5b3c1151fb8f5f13

joey-grafana (approved):

PR #837, opened 2026-08-04
fix(deps): resolve pnpm audit vulnerabilities

shelldandy:
## Summary

Bumps `pnpm-workspace.yaml` overrides to remediate `pnpm audit` findings. Audit went from **8 findings (4 high / 4 moderate)** to **4 findings (1 high / 3 moderate)**; every remaining item has no installable patch (details below).

| Package | Before | After | Advisory |
| --- | --- | --- | --- |
| `brace-expansion@^5` | `>=5.0.8` | `>=5.0.9` | GHSA-rgw5-rvv9-x895 |
| `brace-expansion@^1` | *(absent)* | `>=1.1.18 <2` | GHSA-mh99-v99m-4gvg, GHSA-rgw5-rvv9-x895 |
| `postcss` | `^8.5.18` | `>=8.5.23` | GHSA-fxqj-rqcc-2cmp |

Notes on the two non-obvious choices:

- **`brace-expansion@^1` needs the `<2` upper bound.** A 1.x copy really is in the tree: `minimatch@3.1.5` declares `brace-expansion: ^1.1.7`. Because pnpm matches the `@^1` selector against that *declared* range, an unbounded `>=1.1.18` happily resolved to **5.0.8** and left the advisory unfixed. The `<2` bound pins it to the 1.x line. Per-major keys are kept separate deliberately — v5 changed the CJS export shape, so collapsing majors would break `minimatch` 3 vs 10.
- **`postcss` was edited in place, not added.** An existing `postcss: ^8.5.18` override was already present, so adding a second `postcss:` key would have created a duplicate YAML key where the last one silently wins.

`fast-uri` keeps a single bare key (no version-scoped split): only a 4.x line exists in this tree, so there is no 3.x branch needing its own key.

## Known remaining

All 4 remaining findings are unfixable today; no override was forced to reach zero.

- **`fast-uri` (high, GHSA-7p8r-x3mc-p8w7)** — the patch `4.1.2` is the newest 4.x and the only version outside the vulnerable range (`>=4.0.0 <4.1.2`), but it was published 2026-07-31 and is rejected by this repo's `minimumReleaseAge: 7200` (5 days): `ERR_PNPM_NO_MATURE_MATCHING_VERSION`. Left at `>=4.1.1` rather than disabling the supply-chain policy or inventing a `minimumReleaseAgeExclude` entry (no such precedent exists in this file). **Re-runnable from ~2026-08-05**, when 4.1.2 ages past the cutoff.
- **`react-router` (moderate x2, GHSA-wrjc-x8rr-h8h6 / GHSA-337j-9hxr-rhxg)** and **`react-router-dom` (moderate, GHSA-jjmj-jmhj-qwj2)** — left at 6.30.4.
  - `react-router-dom-v5-compat@6.30.4` is in the tree (via `@grafana/ui@13.1.0`) and its `dist/index.js` imports `UNSAFE_useRoutesImpl` and `UNSAFE_useRouteId` directly from `react-router`, calling them at lines 605 and 1056. Loading `react-router@7.18.0` shows both are `undefined` (as is `UNSAFE_logV6DeprecationWarnings`), so bumping to the patched `>=7.18.0` breaks v5-compat at runtime.
  - Worth noting the usual "React 18 blocks v7" reasoning does *not* apply here: `react-router@7.18.0` peers are `react: >=18`, which this repo (React 18.3.1) satisfies. The blocker is purely the removed `UNSAFE_*` exports. (`react-router@8.x` peers `react: >=19.2.7` and is separately unusable.)
  - `react-router-dom`'s patch `6.30.5` was **never published** — 6.x dead-ends at the flagged 6.30.4.

Two advisories mentioned in the original remediation plan (`GHSA-v2hh-gcrm-f6hx` for fast-uri, `GHSA-r28c-9q8g-f849` for postcss) do not fire in this tree.

## Test plan

- [x] `pnpm install --no-frozen-lockfile` succeeds with all supply-chain policies (`minimumReleaseAge`, `trustPolicy: no-downgrade`, `blockExoticSubdeps`, `strictDepBuilds`) left enabled
- [x] `pnpm audit` — 8 findings (4 high / 4 moderate) to 4 (1 high / 3 moderate)
- [x] `pnpm run typecheck` (`tsc --noEmit`) passes
- [x] `pnpm run build` (production webpack) passes; only a pre-existing asset-size warning
- [x] Verified overrides are not inert by listing `node_modules/.pnpm/` and re-checking the lockfile: `brace-expansion` resolves to exactly `1.1.18` + `5.0.9` (no 5.0.8 references remain), `postcss` to a single `8.5.25`
- [x] Verified the major split preserves export shapes: `minimatch@3.1.5` still loads as a callable function, `minimatch@10.2.5` as a namespace, both matching correctly

Only `pnpm-workspace.yaml` and `pnpm-lock.yaml` are touched; no audit suppressions were added.

### Note for reviewers: `postcss` major bound

Main's override was `postcss: ^8.5.18`, bounded to 8.x; this PR uses `postcss: '>=8.5.23'`, which is unbounded upward. That matches the advisory's patched range and resolves cleanly today (single `8.5.25`; no mature 9.x on npm). Flagging it because an unbounded `>=` selector is structurally the same footgun that made the `brace-expansion@^1` key resolve to 5.0.8 — it can drift across majors once a new major publishes. Happy to switch to `'>=8.5.23 <9'` to preserve main's 8.x ceiling if preferred.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -27,9 +27,10 @@ overrides:
   js-yaml: 4.3.0
   '@opentelemetry/core': 2.8.0
   qs: ^6.15.2
-  brace-expansion@^5: '>=5.0.8'
+  brace-expansion@^1: '>=1.1.18 <2'
+  brace-expansion@^5: '>=5.0.9'
   fast-uri: '>=4.1.1'
-  postcss: ^8.5.18
+  postcss: '>=8.5.23'
 
 importers:
 
@@ -2589,11 +2590,11 @@ packages:
   body@5.1.0:
     resolution: {integrity: sha512-chUsBxGRtuElD6fmw1gHLpvnKdVLK302peeFa9ZqAEk8TyzZ3fygLyUEDDPTJvL9+Bor0dIwn6ePOsRM2y0zQQ==}
 
-  brace-expansion@1.1.16:
-    resolution: {integrity: sha512-IDw48K2/2kRkg9LdJxurvq3lV3aBgq0REY89duEqFRthjlPdXHKMj7EnQOXVckxzgisinf3nHfrcE2FufFLXMw==}
+  brace-expansion@1.1.18:
+    resolution: {integrity: sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==}
 
-  brace-expansion@5.0.8:
-    resolution: {integrity: sha512-JZyDyq3D4AUifKTPOB7DELf6XsB3WdPuNxCtob1vFXPsSXhdAiHBWJ/tJ8HAc9aH84BK+5JFZLNkJKx3G9kzQg==}
+  brace-expansion@5.0.9:
+    resolution: {integrity: sha512-ScQ4IuvIEF1TMlP7Zt+vjJ//9zlPb2SDcxWxM3bk8s6t6GGdJ7KO1dCcTidOPJKePW30LE/2cT7wCyPho9/Wxg==}
     engines: {node: 20 || >=22}
 
   braces@3.0.3:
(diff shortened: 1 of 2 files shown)


shelldandy:
Superseded by #852, which rebases onto current `main` and carries these overrides forward (plus this run's new CVE fixes). This branch is `CONFLICTING` against `main` and its unique changes have since landed upstream, so closing to avoid a duplicate/regressing merge. Reopen if I've missed something.

PR #846, opened 2026-08-10
fix(seeker): keep resize handles updating range

hugohaggmark:
**What is this feature?**

Follows up on #834. That fix stopped a refresh from turning a relative range into an absolute one, but it also broke the time seeker's left and right resize handles, since resizing the selection doesn't fire a cursor `MouseEvent`. This keeps the original selection handling so drag and resize both work, and instead flags the programmatic brush redraw with `isProgrammaticSelect` so only that redraw is ignored.

So the seeker's resize handles update the time range again, without bringing back #815.

For #834

This only affects users with the `tracesDrilldownTimeSeeker` feature toggle enabled, since the time seeker is behind it.

I'm new to this codebase, so I've tried to cover the cases manually (refresh keeps the relative range, and drag and resize both update it), but someone with more experience should test it out. 🤞

I dropped testUtils and the component tests I added, it needed too much uPlot stubbing to be readable. A proper test for this belongs in `e2e/time-seeker.spec.ts`, which is commented out until Grafana 13+ is released.


src/components/Explore/seeker/useTimeSeekerChartConfig.ts
```diff
@@ -100,20 +99,19 @@ export function useTimeSeekerChartConfig({
...
-      const newRange: AbsoluteTimeRange = { from, to };
-      setTimelineRange(newRange);
+        if (!suppressNextTimeRangeUpdate.current) {
+          onChangeTimeRange(newRange);
+        }
 
-      if (!suppressNextTimeRangeUpdate.current) {
-        onChangeTimeRange(newRange);
+        suppressNextTimeRangeUpdate.current = false;
       }
```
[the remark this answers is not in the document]
hugohaggmark: I reverted all the changes back to d4456391e7da4a67f794bcccf08442bf34478094 and added just this line 
```ts
isProgrammaticSelect.current = true;
``` 
to the `requestAnimationFrame` callback to minimize more regressions.

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
```diff
@@ -100,20 +99,19 @@ export function useTimeSeekerChartConfig({
...
-
-      suppressNextTimeRangeUpdate.current = false;
     });
```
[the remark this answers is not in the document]
hugohaggmark: I reverted all the changes back to d4456391e7da4a67f794bcccf08442bf34478094 and added just this line 
```ts
isProgrammaticSelect.current = true;
``` 
to the `requestAnimationFrame` callback to minimize more regressions.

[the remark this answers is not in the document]
hugohaggmark: proper test should be added to e2e tests instead

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
```diff
@@ -100,20 +99,19 @@ export function useTimeSeekerChartConfig({
         return;
       }
 
-      if (!isUserSelection(u)) {
-        return;
-      }
+      const xDrag = Boolean(u.cursor?.drag?.x);
+      if (xDrag && u.select.left != null && u.select.width != null) {
```
joey-grafana: This treats width: 0 as a real selection

We fixed that in the last PR so just wondering if it's not needed here :)

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
```diff
@@ -153,6 +151,7 @@ export function useTimeSeekerChartConfig({
       requestAnimationFrame(() => {
         const left = u.valToPos(timelineRange.from, 'x');
         const right = u.valToPos(timelineRange.to, 'x');
+        isProgrammaticSelect.current = true;
```
joey-grafana: Should we set this in other places before setSelect is called?

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
hugohaggmark: I didn't want to introduce a new regression again but I can have a look 👍

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
hugohaggmark: yeah, I'll take some time to investigate

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
hugohaggmark: fixed in a29312d6368c9ace738d05ae2298610d46892e54 👍

src/components/Explore/seeker/useTimeSeekerChartConfig.ts
hugohaggmark: changed the solution completely in a29312d6368c9ace738d05ae2298610d46892e54, please take a look @joey-grafana 🙏

joey-grafana (approved):
Nice, thank you!

Issue #839, opened 2026-08-05
Add query-and-count existence check before showing cross-app links


shelldandy:
## Why

metrics-drilldown has the strongest existence-check pattern in the estate, and it is worth copying here rather than redesigning. It is two-layered:

1. **Connector-level gating** — verifies via `getTagKeys`/`getTagValues` that every filter key and value actually exists in the target datasource (`src/Integrations/logs/labelsCrossReference.ts:31-78`)
2. **Query-and-count** — includes a datasource only when `rowCount > 0` (`RelatedLogsOrchestrator.ts:156-190`, `countLogsLines` at `:203-205`)

One nuance when reusing it: the "Related logs" tab there is rendered *unconditionally* (`MetricActionBar.tsx:155-160`) — the check gates the tab's **content and count badge**, not its visibility. Decide deliberately which behaviour is wanted here.

## Done when

- [x] Outbound cross-app links are shown only when the target actually has matching data
- [x] Check covers both key/value existence and a non-zero result count
- [x] Empty-state behaviour is deliberate (hide the link vs. show a disabled/empty tab)

---
*Part of the OTel native support work across the Drilldown apps. Tracked in the `grafana/drilldown` epic (Grafana internal).*

matyax:
If I understand this issue correctly, it should be done by:

- https://github.com/grafana/grafana/pull/128702
- https://github.com/grafana/grafana/pull/130515

@joey-grafana / @shelldandy  feel free to reopen otherwise or if there's anything missing.

Issue #780, opened 2026-06-12
Suggestion: Introduce "Logs for this trace" button


rafaelroquetto:
Analogous to "Logs for this span", a "Logs for this trace" button could provide value, as a given span may not necessarily have a corresponding log line enriched with its value - but the surrounding log lines with the same trace-id can provide valuable context.

It's important to bear in mind though, that unlike "Logs for this span", a "Logs for this trace" button could return results that span more than one service.

In my very humble opinion, such a button, combined with a functionality in which either this button itself or "Logs for this span" appear disabled when there are not matching logs to be found, would provide for a better user experience.

Feel free to reach out if you have any questions.

Thanks!

Issue #550, opened 2025-09-05
Open Logs Drilldown when clicking on "logs for this span" in Traces Drilldown


ifrost:
Use https://github.com/grafana/grafana/issues/110681 to provide custom behavior to a data link "logs for this span" and open Logs Drilldown instead of Classic Explore.

Was a subtask of https://github.com/grafana/asserts-app-plugin/issues/1933 and has been extracted to a separate issue.

POC:
- [Logs Drilldown](https://github.com/grafana/logs-drilldown/compare/ifrost/custom-data-links-behavior?expand=1): Exposing a function to get a link to the app (we cannot use extensions directly inside data links post processors)
- [Traces Drilldown](https://github.com/grafana/traces-drilldown/compare/ifrost/custom-data-links-behavior?expand=1): Add custom behavior

PR #847, opened 2026-08-10
feat(traces-to-logs): resolve logs targets in layers with a count gate

### LINKED ISSUE #779: "Logs for this span" does not filter by trace_id, shows all logs for the service instead
## Description

When clicking "Logs for this span" on a span in the Traces Drilldown app, the resulting Loki query only filters by `service_name` (and `service_namespace`), without including a `trace_id` filter. This causes all logs from the service to be shown instead of only the logs correlated with the selected span/trace.

## Steps to Reproduce

1. Open Traces Drilldown app
2. Select a trace
3. Select a span
4. Click "Logs for this span"

## Expected Behavior

The Loki query should include a `trace_id` filter, e.g.:
```logql
{`service_name`="frontend", `service_namespace`="demo"} | `trace_id` = "<`trace_id`>"
```

## Actual Behavior

The generated Loki query only filters by service labels:
```logql
{`service_name`="frontend", `service_namespace`="demo"}
```

No `trace_id` filter is applied, resulting in all logs from the service being shown.

## Additional Context

- `trace_id` is stored as Loki structured metadata (not a stream label), which is why `{trace_id="..."}` cannot be used as a stream selector. A label filter (`| trace_id = "..."`) is required instead.
- The Tempo datasource "Trace to logs" custom query configuration is ignored by the Traces Drilldown app.
- Grafana version: 13.1.0 (Cloud Advanced)
- Traces Drilldown app: grafana-exploretraces-app

cedricziel:
### ✨ Description

Fixes: #779
Related issue(s): #780, #839, #550

Trace to logs only rendered when `tracesToLogsV2.datasourceUid` was set on the Tempo data source (`createSpanLink.tsx:159`). Unset → no link. Set without `filterByTraceID` → link to every log line for the service (#779).

**Layered resolution** (`src/tracesToLogs/`), precedence order:

| Layer | Source | Owns span links |
|---|---|---|
| Configured | `tracesToLogsV2` on Tempo DS, Loki target | No — core renders them |
| Configured, non-Loki target | Splunk, Elastic, … | Nothing at all — out of scope, left to core |
| Configured, no trace filter | same, `filterByTraceID` off | Yes — core's link is unfiltered, ours is trace-scoped |
| Detected | probe ≤5 Loki DS, default first | Yes |

**Query shapes**, probed in order, first with rows wins:

| Id | Selector + filter |
|---|---|
| `otel-structured-metadata` | `{service_name=~…} \| trace_id="…"` |
| `service-parsed` | `… \| logfmt \| json \| drop __error__ \| trace_id="…"` |
| `otlp-gateway-json` | `{exporter="OTLP", job=~"(.*/)?…"} \| json \| traceid="…"` |
| `job-parsed` | `{job=~"(.*/)?…"} \| logfmt \| json \| trace_id="…"` |
| `line-contains` | `{service_name=~…} \|= "…"` |

**Gating**: link shown only when a probe returned rows (#839), over the trace's own padded window. Per-span query = service + trace id (#779). Trace-wide action covers spans with no own log line (#780).

**States**: spinner while probing → live link → disabled + reason. Never absent (#780).

**Cost**: the resolved pair (`tempoUid → {datasourceUid, strategyId}`) is remembered in localStorage and tried first, so the steady state is one query per trace. A stale pair misses and full probing resumes. Nothing is written to any shared resource; where the config exists but lacks a trace filter, the tooltip says so and links to the data source settings.

#### Decisions

- Links injected into the trace data frame, not via `createSpanLink`. That option replaces core's whole factory (`TraceView.tsx:154`) — trace-to-metrics, trace-to-profiles, span references, session links all lost.
- Panel build held until resolution settles (cap 2.5s); links written to frames in place before the panel exists. Re-emitting data clears every open span detail (`useDetailState.ts:14`); Scenes replaces frame objects on reprocess. Past cap: render without span links rather than collapse the user's open span.
- Span link titled "Trace-filtered logs" — core hardcodes "Logs for this span" (`SpanDetailLinkButtons.tsx:75`).
- Probe count only, no `getTagKeys`/`getTagValues` pre-check: cheaper and exact for a trace-id filter.
- Loki only. Detecting Splunk/Elastic/etc. means maintaining five more query dialects; those stay core's job via `tracesToLogsV2`.
- Fallback span links open Explore, not Logs Drilldown — `getFieldLinksForExplore` gets no `DataLinksContext` post-processor. Trace-wide action does use the extension point. Follow-up to #550.

- [x] This pull request was substantially generated with AI assistance ([GenAI policy](https://github.com/grafana/traces-drilldown/blob/main/docs/genai.md))

### 🧪 How to test?

`pnpm test:ci` 494 pass (37 in `src/tracesToLogs/`), `pnpm typecheck`, `pnpm lint` clean.

Manual, run against the devenv from #848 (Tempo + Loki + correlated log generator):

| Case | Result |
|---|---|
| Logs exist, no DS config | span links + live action, both trace-filtered | ✅ |
| No logs for trace | disabled action, no span links | ✅ |
| Each of 5 shapes vs real Loki | matches own flavour only | ✅ |
| Reopen once configured | core renders links, app adds none | ✅ |
| Config without `filterByTraceID` | core's unfiltered link + our trace-filtered one + repair action | ✅ |
| Open span detail during resolution | stays open | ✅ |


cedricziel:
Stacked follow-up in #848: adds Loki plus a trace-and-log generator to the devenv, E2E coverage, and turns the empty case into a disabled button with a reason rather than rendering nothing.

That PR closes out the manual test plan listed here. All six scenarios were run against the devenv and pass, including write back and the configured-wins precedence check. The one behavioural change it makes to this PR's code is the disabled state, per #780.

Issue #800, opened 2026-06-25
Ensure back button works for main actions


joey-grafana:
Go over how the browser back button operates and make sure it works for all of the main actions such as going back from adding a filter, switching tabs etc

- [x] Ensure back and forward navigation works when the user selects a RED Panel https://github.com/grafana/traces-drilldown/pull/849
- [x] Ensure back and forward navigation works when the user selects include / exclude filter https://github.com/grafana/traces-drilldown/pull/853
- [x]  Ensure back and forward navigation works when the user selects tabs https://github.com/grafana/traces-drilldown/pull/866
- [ ]  Ensure back and forward navigation works when the user adds / removes / edits an AdHoc filter (known issue in Scenes https://github.com/grafana/scenes/issues/1103)

hugohaggmark:
Not sure how much more we can do on this issue as the next big main action is the filtering component that already has an issue waiting to be picked up. I think I'm going to close this issue for now but we can always reopen it if we find more main flows that need improving 👍

PR #849, opened 2026-08-11
fix: stop duplicate history entries on load

hugohaggmark:
**What is this feature?**

Fixes so the group-by default no longer pollutes the browser history on load. The default-seed effect in `AttributesBreakdownScene` used to write `var-groupBy` through the user-action path, which pushed history entries. Combined with `favoriteAttributes` loading async, this created two junk entries on every load (`var-groupBy` walked `resource.service.name` -> `undefined` -> `resource.service.name`), so the back button did nothing useful until you'd pressed it twice.

The seed now passes `ignore` so it uses `replaceState` instead of `pushState`:

```ts
useEffect(() => {
  if (!groupBy || groupBy === 'All' || groupBy === '') {
    model.onChange(favoriteAttributes[0], true);
  }
}, [groupBy]);
```

Also adds `data-testid` to `REDPanel` and `MiniREDPanel` (via a new `testIds.redPanel(metric)` helper) and back-button navigation e2e tests in `e2e/navigation.spec.ts`.

So the back button behaves for a user from the moment the page loads.

For https://github.com/grafana/traces-drilldown/issues/800

This is the first of several fixes for #800, so it intentionally does not close the issue. Other main actions (tab switch, filter add, opening a trace, time range) still misbehave with the back button and will follow in separate PRs.


src/utils/testIds.ts
```diff
@@ -2,4 +2,5 @@ export const testIds = {
   emptyState: 'data-testid empty-state',
   errorState: 'data-testid error-state',
   loadingState: 'data-testid loading-state',
+  redPanel: (metric: string) => `data-testid ${metric}-panel`,
```
hugohaggmark: maybe this should be static entries only and I could add a helper function that I could use throughout the changes instead 🤔

e2e/navigation.spec.ts
```diff
@@ -19,3 +21,81 @@ test.describe('navigating app', () => {
...
+});
+
+type MetricType = 'rate' | 'errors' | 'duration';
+
+async function assertBackAndForwardNabigationWorks(page: Page, startMetric: MetricType, switchToMetric: MetricType) {
+  const explorePage = new ExplorePage(page);
+  await explorePage.assertNotLoading();
+
+  assertREDPanelRadioVisible(page, startMetric);
+  assertREDPanelRadioVisible(page, switchToMetric);
```
joey-grafana: Should we also add await here and other places?

e2e/navigation.spec.ts
```diff
@@ -19,3 +21,81 @@ test.describe('navigating app', () => {
...
+}
+
+async function assertREDPanelRadioVisible(page: Page, metric: MetricType) {
+  await expect(page.getByTestId(testIds.redPanel(metric))).toBeVisible();
+  await expect(page.getByTestId(testIds.redPanel(metric)).getByRole('radio')).toBeVisible();
+}
+
+async function assertCheckedForREDPanelRadio(page: Page, metric: MetricType) {
+  const isChecked = await page.getByTestId(testIds.redPanel(metric)).getByRole('radio').isChecked();
+  return expect(isChecked, `expected radiobutton '${metric}' to be checked`).toBe(true);
```
joey-grafana: isChecked() is a one-shot read — after click/back/forward the radio may not have updated yet. Prefer await expect(locator).toBeChecked() / .not.toBeChecked() so Playwright retries, and await the helper at call sites.

src/components/Explore/TracesByService/Tabs/Breakdown/AttributesBreakdownScene.tsx
```diff
@@ -164,7 +164,7 @@ export class AttributesBreakdownScene extends SceneObjectBase<AttributesBreakdow
 
     useEffect(() => {
       if (!groupBy || groupBy === 'All' || groupBy === '') {
-        model.onChange(favoriteAttributes[0]);
+        model.onChange(favoriteAttributes[0], true);
```
joey-grafana, line 171: ignore: true skips history but still fires breakdown_group_by_changed on load. That inflates “user” metrics. Could also ignore `reportAppInteraction` in `onChange` in this case.

e2e/navigation.spec.ts
hugohaggmark: great catch, I'll fix 🙈

e2e/navigation.spec.ts
hugohaggmark: I started using `toBeChecked() / .not.toBeChecked()` but unfortunately `toBeChecked()` worked but  the  `.not.toBeChecked()` didn't work 🤔 for some reason

src/components/Explore/TracesByService/Tabs/Breakdown/AttributesBreakdownScene.tsx
hugohaggmark: great catch, I'll have a look 👍

e2e/navigation.spec.ts
hugohaggmark: fixed in 24aa8f1824c824bf2306c17c1f1e62116301e896 👍

e2e/navigation.spec.ts
hugohaggmark: I wasn't able to use toBeChecked here because of how the panels re-render so the retry fails. Fixed in 24aa8f1824c824bf2306c17c1f1e62116301e896 using `toHaveAttribute` instead that is also retriable 👍

src/components/Explore/TracesByService/Tabs/Breakdown/AttributesBreakdownScene.tsx
hugohaggmark: Fixed in 24aa8f1824c824bf2306c17c1f1e62116301e896 👍

e2e/navigation.spec.ts
```diff
@@ -1,9 +1,11 @@
 import { expect, test } from '@grafana/plugin-e2e';
 import { ExplorePage } from './fixtures/explore';
+import { testIds } from '../src/utils/testIds';
+import { Locator, Page } from '@playwright/test';
```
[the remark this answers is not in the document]
hugohaggmark: Fixed in 24aa8f1824c824bf2306c17c1f1e62116301e896 👍

e2e/navigation.spec.ts
```diff
@@ -19,3 +21,81 @@ test.describe('navigating app', () => {
...
+  });
+
+  test('clicking on duration panel, browser back and browser forward should work as expected', async ({ page }) => {
+    await assertBackAndForwardNabigationWorks(page, 'rate', 'duration');
+  });
+});
+
+type MetricType = 'rate' | 'errors' | 'duration';
+
+async function assertBackAndForwardNabigationWorks(page: Page, startMetric: MetricType, switchToMetric: MetricType) {
```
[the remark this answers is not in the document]
hugohaggmark: Fixed in 24aa8f1824c824bf2306c17c1f1e62116301e896 👍

e2e/navigation.spec.ts
```diff
@@ -19,3 +21,81 @@ test.describe('navigating app', () => {
...
+  const explorePage = new ExplorePage(page);
+  await explorePage.assertNotLoading();
+
+  assertREDPanelRadioVisible(page, startMetric);
+  assertREDPanelRadioVisible(page, switchToMetric);
+
+  assertCheckedForREDPanelRadio(page, startMetric);
+  assertUnCheckedForREDPanelRadio(page, switchToMetric);
+
```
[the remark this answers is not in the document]
hugohaggmark: Fixed in 24aa8f1824c824bf2306c17c1f1e62116301e896 👍

src/components/Explore/TracesByService/MiniREDPanel.tsx
```diff
@@ -158,7 +168,11 @@ export class MiniREDPanel extends SceneObjectBase<MiniREDPanelState> {
     }
 
     return (
-      <div className={css([styles.container, styles.clickable])} onClick={() => selectMetric(embeddedMini)}>
+      <div
+        className={css([styles.container, styles.clickable])}
+        onClick={(e) => selectMetric(embeddedMini)}
```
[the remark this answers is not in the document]
hugohaggmark: Fixed in 24aa8f1824c824bf2306c17c1f1e62116301e896 👍

joey-grafana (approved):

PR #853, opened 2026-08-13
fix: ignore attribute changes that are not human

hugohaggmark:
**What is this feature?**

Fixes so clicking include/exclude no longer pollutes the browser history. When that attribute gets added to the filters, `AttributesSidebar` auto-advances the selection to the next attribute in a `useEffect`. That auto-advance went through the same `onAttributeChange` path as a real user click, so it pushed its own history entry on top of the filter-click entry, breaking back/forward for include and exclude. This affects both the Breakdown and Comparison tabs, since both render `AttributesSidebar`.

The auto-advance now passes `ignore`, same pattern as #849:

```ts
if (nextIndex < filteredAttributes.length) {
  onAttributeChange(filteredAttributes[nextIndex].value, true);
  return;
}
```

`onChange` uses `ignore` to skip the `reportAppInteraction` call and use `replaceState` instead of `pushState`, so only the actual filter click lands in history.

Also adds `data-testid`/`data-selected` to sidebar attribute items (via a new `getTestIdFromAttribute` helper), unit tests pinning the `ignore` mechanism on both scenes, include/exclude back-button e2e tests in `e2e/navigation.spec.ts`, and an `e2e:fast:10x` command so you can run the new e2e tests 10 times to check for flakiness.

So the back button behaves correctly when a user includes or excludes an attribute in the filter.

For https://github.com/grafana/traces-drilldown/issues/800

This is another step on #800 (following #849), so it intentionally does not close the issue. Tab switching and time range still need their own fixes.

### DIFF
--- a/e2e/navigation.spec.ts
+++ b/e2e/navigation.spec.ts
@@ -1,7 +1,9 @@
 import { expect, test } from '@grafana/plugin-e2e';
 import { ExplorePage } from './fixtures/explore';
-import { getTestIdFromMetric, testIds } from '../src/utils/testIds';
+import { getTestIdFromAttribute, getTestIdFromMetric, testIds } from '../src/utils/testIds';
 import type { Page } from '@playwright/test';
+import { Components } from '@grafana/e2e-selectors';
+import { AttributeItem } from '../src/types';
 
 test.describe('navigating app', () => {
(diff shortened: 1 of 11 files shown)


joey-grafana (approved):
Nice, works as expected for include/exclude 👍

PR #854, opened 2026-08-13
feat: Add TraceView resource attribute links

joey-grafana:
### ✨ Description

Part of https://github.com/grafana/grafana/issues/129072

Similar to https://github.com/grafana/asserts-app-plugin/pull/3682

Adds TraceView links on resource attributes (grafana/traceview/resource-attributes) that open Traces Drilldown with matching filters.

### DIFF
--- a/src/utils/resourceAttributes.test.ts
+++ b/src/utils/resourceAttributes.test.ts
@@ -1,0 +1,137 @@
+import { type AbsoluteTimeRange, type PluginExtensionResourceAttributesContext } from '@grafana/data';
+
+import pluginJson from '../plugin.json';
+import {
+  makeTraceResourceAttributeLink,
+  RESOURCE_ATTRIBUTE_LINK_ICON,
+  RESOURCE_ATTRIBUTE_LINK_LABEL,
+  TRACE_RESOURCE_ATTRIBUTE_LINKS,
+} from './resourceAttributes';
+
+const timeRange: AbsoluteTimeRange = { from: 1000, to: 2000 };
+
+const context: PluginExtensionResourceAttributesContext = {
+  datasource: { uid: 'tempo', type: 'tempo' },
+  attributes: {
+    'service.name': ['cart'],
+    'service.version': ['1.2.3'],
+    'k8s.container.name': ['cart'],
+    'k8s.deployment.name': ['cart-deploy'],
+  },
+  timeRange,
+};
+
+function linkFor(attributeName: string) {
+  const config = TRACE_RESOURCE_ATTRIBUTE_LINKS.find((c) => c.attributeName === attributeName);
+  if (!config) {
+    throw new Error(`Missing TRACE_RESOURCE_ATTRIBUTE_LINKS entry for ${attributeName}`);
+  }
+  return makeTraceResourceAttributeLink(config);
+}
+
+function configure(attributeName: string, overrides?: Partial<PluginExtensionResourceAttributesContext>) {
+  return linkFor(attributeName).configure?.({ ...context, ...overrides });
+}
+
+function drilldownUrl(path: string) {
+  return new URL(path, 'https://example.com');
+}
+
+describe('TRACE_RESOURCE_ATTRIBUTE_LINKS', () => {
+  it('registers a unique TraceView category per attribute', () => {
+    const categories = TRACE_RESOURCE_ATTRIBUTE_LINKS.map((c) => c.attributeName);
+    expect(new Set(categories).size).toBe(categories.length);
+  });
+
+  it('declares matching title and description in plugin.json', () => {
+    expect(pluginJson.extensions?.addedLinks).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          targets: ['grafana/traceview/resource-attributes'],
+          title: RESOURCE_ATTRIBUTE_LINK_LABEL,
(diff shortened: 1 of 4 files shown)


src/utils/resourceAttributes.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+function drilldownUrl(path: string) {
+  return new URL(path, 'https://example.com');
+}
+
+describe('TRACE_RESOURCE_ATTRIBUTE_LINKS', () => {
+  it('registers a unique TraceView category per attribute', () => {
+    const categories = TRACE_RESOURCE_ATTRIBUTE_LINKS.map((c) => c.attributeName);
+    expect(new Set(categories).size).toBe(categories.length);
+  });
```
hugohaggmark: nit: this might be a good candidate for a table driven test?

src/utils/resourceAttributes.ts
```diff
@@ -0,0 +1,147 @@
...
+export type TraceResourceAttributeLinkConfig = {
+  /** TraceView row key (`link.category === attribute.key`). Always required. */
+  attributeName: string;
+  /**
+   * Resource attributes applied as TraceQL filters.
+   * Defaults to `[attributeName]`. Omit the row key when a secondary attribute
+   * should open a parent entity (e.g. container.name → k8s.deployment.name).
+   */
+  filters?: string[];
+};
```
hugohaggmark: nit: do we generally prefer type over interface for simple type declarations like this? I don't have a strong preference either way, just curious about the convention here.

src/utils/resourceAttributes.ts
```diff
@@ -0,0 +1,147 @@
...
+  type PluginExtensionAddedLinkConfig,
+  PluginExtensionPoints,
+  type PluginExtensionResourceAttributesContext,
+} from '@grafana/data';
+
+import { EXPLORATIONS_ROUTE, RESOURCE_ATTR, VAR_DATASOURCE, VAR_FILTERS, VAR_PRIMARY_SIGNAL } from './shared';
+
+export type TraceResourceAttributeLinkConfig = {
+  /** TraceView row key (`link.category === attribute.key`). Always required. */
+  attributeName: string;
```
hugohaggmark: could we narrow the type down from just string here, or are the attribute names dynamic?

src/utils/resourceAttributes.ts
```diff
@@ -0,0 +1,147 @@
...
+  config: TraceResourceAttributeLinkConfig
+): PluginExtensionAddedLinkConfig<PluginExtensionResourceAttributesContext> {
+  const filters = filterNames(config);
+  const required = [...new Set([config.attributeName, ...filters])];
+
+  return {
+    targets: [PluginExtensionPoints.TraceViewResourceAttributes],
+    ...linkCopy,
+    // TraceView matches the link to an attribute row via category === attribute.key
+    category: config.attributeName,
```
hugohaggmark, line 109: I think category is deprecated in favor of group, if I remember correctly. I believe group was introduced in this PR: https://github.com/grafana/grafana/pull/116481

src/utils/resourceAttributes.ts
```diff
@@ -0,0 +1,147 @@
...
+  icon: RESOURCE_ATTRIBUTE_LINK_ICON,
+};
+
+function attrValue(attributes: Record<string, string[]> | undefined, name: string): string | undefined {
+  return attributes?.[name]?.[0]?.trim() || undefined;
+}
+
+/** filterNames — TraceQL filter keys; defaults to the row attribute */
+function filterNames(config: TraceResourceAttributeLinkConfig): string[] {
+  return config.filters ?? [config.attributeName];
```
hugohaggmark: should we also do this when filters is an empty array?

hugohaggmark (dismissed):
Looks good 🎉  but I left some comments for you to consider 👍

src/utils/resourceAttributes.test.ts
joey-grafana: Sounds good - switched this to it.each so each attribute is its own case and a duplicate shows up as that name rather than a set-size mismatch.

src/utils/resourceAttributes.ts
joey-grafana: We usually use types for similar kinds of simple type declarations

src/utils/resourceAttributes.ts
joey-grafana: Narrowed!

src/utils/resourceAttributes.ts
joey-grafana: Added in https://github.com/grafana/grafana/pull/130684

src/utils/resourceAttributes.ts
joey-grafana: Good catch. ?? wouldn’t cover []. We now default to the row attribute when filters is omitted or empty, so we never open drilldown with no filters.

src/utils/resourceAttributes.ts
hugohaggmark: Awesome, thanks for sharing 👍

src/utils/resourceAttributes.ts
hugohaggmark: Aah of course, thanks 👍

hugohaggmark (approved):
Awesome work 🎉

PR #855, opened 2026-08-14
chore(e2e): introduce tracesExplore fixture

hugohaggmark:
**What is this feature?**

Replaces the old `explore` e2e fixture with a proper `tracesExplorePage` fixture backed by a `TracesExplorePage` page object model. The model owns navigation (`goto`), readiness checks (`waitForExploreReady`, `assertNotLoading`) and teardown (`unroute`), so the specs stop repeating that setup inline.

So the e2e specs share one reliable way to reach a ready explore view instead of duplicating navigation and wait logic.

Fixes #


joey-grafana (dismissed):

matyax (approved):
Changes look good!

PR #868, opened 2026-08-21
chore: add cp-update workflow, bump action pins

### LINKED ISSUE #319: Add bundle-size and create-plugin-update workflows
This project is lacking bundle-size and create-plugin-update workflows that can help with maintenance and reviewing frontend bundle size impact in PRs.

Please see each readme to understand how to add them to this project:

- [Bundle Size Readme](https://github.com/grafana/plugin-actions/blob/main/bundle-size/README.md)
- [Create Plugin Update Readme](https://github.com/grafana/plugin-actions/blob/main/create-plugin-update/README.md)

Note that the bundle size workflow expects "main build information" to be stored in gh action artifacts which means you will need to build the plugin on merges to main and then upload the built artifact to gh. [The readme troubleshooting section](https://github.com/grafana/plugin-actions/blob/main/bundle-size/README.md#troubleshooting) should help explain how to implement this.

hugohaggmark:
**What is this feature?**

Adds a `create-plugin-update` workflow and bumps the pins in the existing `bundle-size` workflow. `actions/checkout` goes to `v7.0.1`, `bundle-size` to `v1.1.1`, and `create-plugin-update` lands on `v2.0.2`. Every action is pinned to the immutable release SHA with a `# version (date)` comment so Dependabot stays happy.

So we get automated bundle-size diffs on PRs and monthly create-plugin scaffolding updates.

Fixes #319

Also dropped `contents: write` to `contents: read` in `bundle-stats.yml` since the bundle-size action only needs read. 

How often should we run this? Monthly seems fine to me

### DIFF
--- a/.github/workflows/cp-update.yml
+++ b/.github/workflows/cp-update.yml
@@ -1,0 +1,26 @@
+name: Create Plugin Update
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 1 * *' # run once a month on the 1st day
+
+# To use the default github token with the following elevated permissions make sure to check:
+# **Allow GitHub Actions to create and approve pull requests** in https://github.com/ORG_NAME/REPO_NAME/settings/actions.
+# Alternatively create a fine-grained personal access token for your repository with
+# `contents: read and write` and `pull requests: read and write` and pass it to the action.
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: grafana/plugin-actions/create-plugin-update@b82357e85d512c678416146ca4e9f8672bd108da # create-plugin-update/v2.0.2 (2026-03-09)
+        # Uncomment to use a fine-grained personal access token instead of default github token
+        # (For more info on how to generate the token see https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
+        # with:
+        # Make sure to save the token in your repository secrets
+        # token: $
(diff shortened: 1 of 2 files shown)


.github/workflows/cp-update.yml
```diff
@@ -0,0 +1,26 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: grafana/plugin-actions/create-plugin-update@b82357e85d512c678416146ca4e9f8672bd108da # create-plugin-update/v2.0.2 (2026-03-09)
```
joey-grafana, line 22: create-plugin-update@v2.0.2 requires token and this workflow doesn’t pass one (the commented token: $ isn’t valid). Every run will fail with Input required and not supplied: token.

hugohaggmark: Good catch, I thought this comment was still true for v2.0.2 🤔 

```md
# To use the default github token with the following elevated permissions make sure to check:
# **Allow GitHub Actions to create and approve pull requests** in https://github.com/ORG_NAME/REPO_NAME/settings/actions.
```

joey-grafana: Hmm are you getting an error in the file itself?

Screenshot 2026-08-21 at 13 26 38

hugohaggmark: No sorry I don't get any errors but I'm using Zed so maybe I'm missing a plugin?

hugohaggmark: I've opened this https://github.com/grafana/deployment_tools/pull/696730

hugohaggmark: and updated the workflow to use GATB

joey-grafana (approved):

PR #873, opened 2026-08-24
fix: always show panel menus and use secondary Include/Exclude buttons

joey-grafana:
### ✨ Description

- Always show the breakdown panel kebab (`showMenuAlways`) so header actions do not leave empty hover-only menu space.
- Replace the Include/Exclude segmented control with two small secondary buttons (breakdown tiles, comparison highest-difference, exceptions table).

### 🧪 How to test?

Screenshot 2026-08-24 at 08 41 56


hugohaggmark (dismissed):
Looks great and works as expected! 🙌

One thought: could these buttons benefit from being extracted into a shared component? WDYT?

torkelo (approved):

hugohaggmark (approved):
Looks great and works as expected 🎉

PR #875, opened 2026-08-24
fix: use locationUtil.assureBaseUrl for subpath URLs

### LINKED ISSUE #872: Use locationUtil.assureBaseUrl instead of manual config.appSubUrl handling
We build subpath-aware URLs by manually reading `config.appSubUrl` in a few places across Traces Drilldown. `@grafana/data` ships `locationUtil.assureBaseUrl`, the canonical way to prepend the app's base URL to an absolute path, and we should adopt it consistently so there is one helper for subpath handling instead of hand-rolled `config.appSubUrl ?? ''` concatenation. `assureBaseUrl` also already handles the edge cases such as trailing and leading slashes and absolute versus relative paths.

Hashim1999164:
## Description
Fixes: #872

Explore links, Loki data links, and the OFREP base URL now use locationUtil.assureBaseUrl instead of concatenating config.appSubUrl.

[x] This pull request was substantially generated with AI assistance

## How to test
* Grafana behind a subpath
* Explore menu and span list links keep the subpath
* Loki drilldown data link keeps the subpath
* pnpm lint, typecheck, and the DataLinksCustomContext tests

### DIFF
--- a/src/featureFlags/openFeature.tsx
+++ b/src/featureFlags/openFeature.tsx
@@ -4,6 +4,7 @@ import { OFREPWebProvider } from '@openfeature/ofrep-web-provider';
 import { StandardResolutionReasons } from '@openfeature/core';
 import { OpenFeature, type Provider, type ResolutionDetails } from '@openfeature/web-sdk';
 
+import { locationUtil } from '@grafana/data';
 import { config, logWarning } from '@grafana/runtime';
 
 /** OpenFeature domain for this plugin’s evaluations (OFREP to Grafana’s feature API). */
@@ -14,9 +15,9 @@ function getFeaturesOfrepBaseUrl(): string | undefined {
   if (typeof window === 'undefined') {
     return undefined;
   }
-  const sub = (config.appSubUrl ?? '').replace(/\/$/, '');
-  const path = `${sub}/apis/features.grafana.app/v0alpha1/namespaces/${encodeURIComponent(config.namespace)}`;
-  const pathname = path.startsWith('/') ? path : `/${path}`;
+  const pathname = locationUtil.assureBaseUrl(
+    `/apis/features.grafana.app/v0alpha1/namespaces/${encodeURIComponent(config.namespace)}`
+  );
   return new URL(pathname, window.location.origin).toString();
 }
 
--- a/src/featureFlags/openFeature.tsx
+++ b/src/featureFlags/openFeature.tsx
@@ -15,9 +15,12 @@ function getFeaturesOfrepBaseUrl(): string | undefined {
   if (typeof window === 'undefined') {
     return undefined;
   }
-  const pathname = locationUtil.assureBaseUrl(
-    `/apis/features.grafana.app/v0alpha1/namespaces/${encodeURIComponent(config.namespace)}`
-  );
+  const pathname = locationUtil
+    .assureBaseUrl(
+      `/apis/features.grafana.app/v0alpha1/namespaces/${encodeURIComponent(config.namespace)}`
+    )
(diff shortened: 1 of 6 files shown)


src/pages/Explore/DataLinksCustomContext.test.tsx
hugohaggmark: the config property can be removed here 👍

src/pages/Explore/DataLinksCustomContext.test.tsx
```diff
@@ -84,6 +84,11 @@ beforeEach(() => {
   jest.clearAllMocks();
   capturedContext = null;
   setupAllConditionsMet();
+  locationUtil.initialize({
+    config: { appSubUrl: '/grafana' } as Parameters<typeof locationUtil.initialize>[0]['config'],
```
hugohaggmark: the typings can be simplified here
```suggestion
    config: { appSubUrl: '/grafana' } as GrafanaConfig,
```

src/featureFlags/openFeature.tsx
```diff
@@ -14,9 +15,9 @@ function getFeaturesOfrepBaseUrl(): string | undefined {
   if (typeof window === 'undefined') {
     return undefined;
   }
-  const sub = (config.appSubUrl ?? '').replace(/\/$/, '');
-  const path = `${sub}/apis/features.grafana.app/v0alpha1/namespaces/${encodeURIComponent(config.namespace)}`;
-  const pathname = path.startsWith('/') ? path : `/${path}`;
+  const pathname = locationUtil.assureBaseUrl(
+    `/apis/features.grafana.app/v0alpha1/namespaces/${encodeURIComponent(config.namespace)}`
+  );
```
hugohaggmark: One thing I noticed that I think we should address before approving this PR. The previous code used `replace(/\/$/, '')`, which would remove a trailing slash from `appSubUrl`, so for example `/grafana/` would become `/grafana`.

If you have the time, could you add a few tests that capture the behavior of the existing code before switching to the new implementation? That way we can make sure we're preserving the existing behavior and that all the tests continue to pass with the changes.

Let me know if I can help out in any way, thanks 🙏

hugohaggmark (commented):
Thanks for contributing to Grafana and Traces Drilldown! 🙌

The code looks good overall, but I found a few things that I'd like you to take a look at. Thanks! 🙏

Hashim1999164:
Addressed the review notes.

Removed the unused runtime config mock, typed the locationUtil config as GrafanaConfig, and added tests so appSubUrl with and without a trailing slash both produce the same OFREP path. assureBaseUrl can join `/grafana/` into a double slash, so that path is normalized the same way the old replace did.

src/pages/Explore/DataLinksCustomContext.test.tsx
Hashim1999164: Done. `locationUtil.initialize` now only passes `{ appSubUrl: '/grafana' } as GrafanaConfig`.

src/pages/Explore/DataLinksCustomContext.test.tsx
Hashim1999164: Updated to the simpler `GrafanaConfig` cast you suggested.

src/featureFlags/openFeature.tsx
Hashim1999164: Added coverage for that trailing slash case.

`openFeature.test.tsx` now checks both `/grafana` and `/grafana/` so the OFREP path stays `/grafana/apis/...` and does not pick up a double slash.

hugohaggmark (approved):
Thanks for your contribution 🎉

PR #890, opened 2026-09-08
fix: skip duration filter without latency

### LINKED ISSUE #850: TracesByServiceScene: unguarded latencyThreshold interpolation emits a dangling 'duration > ' before RED histogram resolves (extends #765)
## Summary

`TracesByServiceScene`'s `buildQuery` interpolates `latencyThreshold` into the TraceQL span-list query before the RED-duration-histogram subscriber that's supposed to set it has ever run, producing a dangling `&& duration > ` clause (empty right-hand side) on the **first** query every Traces Drilldown page load fires. This is the same unguarded-Scenes-variable-interpolation class as #765 and #95, just a third variant neither of those (nor `isTransientMalformedTraceQLFailure` in the e2e harness that reconciles the other two) currently recognises.

**Version tested:** plugin `2.1.0`, tag `v2.1.0` = `932bac31b336045371c26739192e3f8967d1606a`. Present continuously since `8b3906f3` (2024-09-05); byte-identical on `main` today.

## Where

`src/components/Explore/TracesByService/TracesByServiceScene.tsx`, `buildQuery`, line 425:

```ts
if (!typeQuery.length) {
  typeQuery = `&& duration > ${VAR_LATENCY_THRESHOLD_EXPR}`;   // line 425
}
...
return {
  refId: 'A',
  query: `{${VAR_FILTERS_EXPR}${typeQuery}}${selectQuery}`,     // line 431
  ...
};
```

`VAR_FILTERS_EXPR = '${primarySignal} && ${filters}'`, `VAR_LATENCY_THRESHOLD_EXPR = '${latencyThreshold}'` (`src/utils/shared.ts:35,41`).

`latencyThreshold` is declared with **no initial value** (`src/pages/Explore/TraceExploration.tsx:529-540`) and is set only from inside the RED duration-histogram's data subscriber (`src/components/Explore/TracesByService/REDPanel.tsx:112-119`). `buildQuery` has no dependency on that resolution, so the span-list query fires first with `${latencyThreshold}` interpolating to the empty string, producing `&& duration > ` with no operand.

## Captured wire evidence

Grabbed over the Grafana Live WebSocket (`ws://.../api/live/ws`), since this datasource is provisioned for streaming search and the malformed query never appears in a plain `/api/ds/query` POST body — which is presumably why this has stayed unnoticed:

```json
{"subscribe":{"channel":"1/ds/cerberus-tempo/search/135ea56e-67bd-4f77-b14e-405b6973aef6",
 "data":{"refId":"A","query":"{span.db.system.name!=\"\" && true&& duration > }",
 "queryType":"traceql","tableType":"spans","limit":200,"spss":10,"filters":[]}},"id":2}
```

Query text: `{span.db.system.name!="" && true&& duration > }` — 47 chars, the trailing `}` at column 47 immediately follows the empty RHS. (The missing space before the second `&&` is the same template's fingerprint as the `errors` arm using `' && status = error'` with a leading space vs. the `duration` arm's `'&& …'` with none.)

The backing Tempo-compatible API correctly rejects it with a `syntax error: unexpected }, expecting an attribute` at that exact column — this is not a backend/datasource bug, it's invalid TraceQL reaching the wire.

## It's not specific to one primary signal

Drove all five enumerated `primarySignal` values; every one fires the dangling query first, then a valid one once the histogram resolves:

| `var-primarySignal` | queries sent, in order | recovers? |
|---|---|---|
| `nestedSetParent<0` (default) | `{nestedSetParent<0 && true&& duration > }` → `… > 134ms` → `… >= 134ms` | yes |
| `true` | `{true && true&& duration > }` → `… > 67ms` → `… >= 67ms` | yes |
| `kind=server` | `{kind=server && true&& duration > }` → `… > 134ms` → `… >= 134ms` | yes |
| `kind=consumer` | `{kind=consumer && true&& duration > }` → `… > 0` → `… >= 0` | yes |
| `span.db.system.name!=""` | `{span.db.system.name!="" && true&& duration > }` | no (unrelated cause: this filter legitimately matches zero spans in our test data, so the histogram never returns and the threshold never resolves — not part of this report) |

So this affects **every page load of Traces Drilldown**, not one specific signal — the signal in the originally-observed report only happened to be the one where the race never self-resolves, making the malformed query the *only* one visible in that session.

## Relation to existing issues

Same class as #765 (`by(undefined)`, open, zero comments) and #95 (closed 2024-12-20, but the referenced fix diff doesn't actually add a guard) — unguarded interpolation of an unresolved Scenes variable straight into a TraceQL string fragment. `iterate-drilldown-apps.spec.ts` in the e2e harness already carries a reconciler (`isTransientMalformedTraceQLFailure`) for two shapes of this; `{… && duration > }` is a third shape it doesn't recognize yet.

A guard for the same general problem already exists elsewhere in this codebase: `src/components/Explore/queries/generateMetricsQuery.ts:36-57` checks a resolved value before emitting the query fragment that consumes it — the same pattern would fix this call site.

## Suggested fix

Either give `VAR_LATENCY_THRESHOLD` an initial non-empty value at `TraceExploration.tsx:531`, or check it's resolved before emitting the `duration >` clause in `buildQuery` (mirroring `generateMetricsQuery.ts:36-57`) and omit the clause entirely until it is. Either fix needs a matching update to `TracesByServiceScene.test.tsx:67`, which currently pins the malformed string as expected output.

## Found via

Discovered while investigating a crawl-oracle finding in [tsouza/cerberus](https://github.com/tsouza/cerberus) (a ClickHouse-backed Prometheus/Loki/Tempo gateway used only as a datasource here) — full investigation notes: https://github.com/tsouza/cerberus/issues/2009

hugohaggmark:
**What is this feature?**

Fixes so the duration query no longer adds `&& duration > ${latencyThreshold}` when no latency threshold is set. `buildQuery` now takes the latency value explicitly and only adds that clause when it's truthy.

So the duration view returns results instead of an empty/invalid comparison when there's no latency filter.

Fixes #850

### DIFF
--- a/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.test.ts
+++ b/src/components/Explore/TracesByService/Tabs/Structure/StructureScene.test.ts
@@ -1,4 +1,4 @@
-import { parseTraces } from './StructureScene';
+import { buildQuery, parseTraces } from './StructureScene';
 import { dumpTree, mergeTraces } from '../../../../../utils/trace-merge/merge';
 import { SearchResponse } from '../../../../../types';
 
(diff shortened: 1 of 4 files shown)


joey-grafana (commented):
Nice this looks good for the trace list but I think we also need this on the structure tab?

hugohaggmark:
> Nice this looks good for the trace list but I think we also need this on the structure tab?

Great catch 🙌 , fixed in https://github.com/grafana/traces-drilldown/pull/890/commits/fd1d4a5f01b02c17dd7f0a35b746045d86b46493 👍

joey-grafana (approved):

PR #892, opened 2026-09-08
fix: initialize i18n before embedded Trace Exploration mounts

joey-grafana:
### ✨ Description

Follow-up to #879 (create-plugin 7.10.0). That bump marked `i18next` as a webpack external while `@grafana/i18n` stays bundled, so `t()` can run against an uninitialized wrapper. The embed path never called `initPluginTranslations` (only the App lazy load did).

- Exposed components (and the App page) now await a shared `initPluginI18n()` before rendering, so `t()` is not called in the `TraceExploration` constructor before translations are ready.
- Fixes `t() was called before i18n was initialized` when Grafana loads Embedded Trace Exploration without opening the full app.

### 🧪 How to test?

- [ ] Open a host that embeds Trace Exploration without visiting Traces Drilldown first — no i18n console error, view loads.
- [ ] Open Traces Drilldown (non-embedded) — should load.

### DIFF
--- a/src/exposedComponents/index.test.tsx
+++ b/src/exposedComponents/index.test.tsx
@@ -1,0 +1,78 @@
+import { dateTime } from '@grafana/data';
+import { render, screen, waitFor } from '@testing-library/react';
+import React from 'react';
+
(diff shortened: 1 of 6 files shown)


hugohaggmark (approved):
Looks good 🎉

PR #904, opened 2026-09-15
fix(deps): align React Router with Grafana runtime

shelldandy:
## Summary
- Keeps React Router on 6.30.6 to match Grafana's externalized v6 runtime
- Preserves independent v5 and v6 router lines required by Grafana compatibility packages
- Documents why the 2 remaining audit findings do not have a compatible package upgrade

## Security rationale
Forcing React Router 7 beneath `react-router-dom` 6.30.6 creates an unsupported mixed-major tree. GHSA-wrjc-x8rr-h8h6 has no compatible v6 package fix, but Traces Drilldown only navigates to a compile-time static internal route. GHSA-337j-9hxr-rhxg affects SSR hydration, which this browser plugin does not use.

## Test plan
- [x] `pnpm run lint`, `pnpm run typecheck`, and `pnpm run build`
- [x] Frozen lockfile install
- [x] Audit contains only the 2 documented moderate advisories

Upstream migration: grafana/grafana#129933

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Pi


joey-grafana (changes requested):
Are we upgrading react-router or leaving it at 6.30.6? Like https://github.com/grafana/metrics-drilldown/pull/1427/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

shelldandy:
> Are we upgrading react-router or leaving it at 6.30.6?

Addressed: we are leaving `react-router` at 6.30.6 and scoping the override to `react-router@^6`, matching `react-router-dom@6.30.6` and Grafana’s externalized v6 runtime. Traces Drilldown’s only React Router destination is the compile-time static internal `ROUTES.Explore`, so no unnecessary application hardening was added. The PR now documents the 2 known remaining advisories.

joey-grafana (approved):

